PR #2, opened 2022-02-02
Tn fixes

tanguyNaets:
- Fix a few bugs
  - Convert gitlab CI .docker script from bash to sh because the alpine `dockerindocker` image in use only has sh installed and it would be more cumbersome to install a bash shell and try to set it as the default shell in the first steps. of the script Also, there are no `dockerindocker` images built on top of a base layer with bash as the default shell.
  - Remove a streamlit flag that does not exist anymore in it's latest poetry compatible version
  - Install poe in fastapi|streamlit app images
- Small performance changes
  - Do not track autogenerated html docs
  - Execute mypy last since it's the slowest linting tool to execute, even when cached
- Update the docs to explain how to remove packages


{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -25,37 +25,41 @@ variables:
...
           {%- endif %}
-          --target $DOCKER_TARGET \
+          --target "$DOCKER_TARGET" \
           --pull \
           $DOCKER_TAGS_JOINED \
           .
+
         # Push all the tagged images.
-        for DOCKER_TAG in ${DOCKER_TAGS[@]}; do
+        for DOCKER_TAG in ${DOCKER_TAGS}; do
```
lsorber, line 62: Thanks for this! I take it these changes convert the existing script from a Bash-dependent syntax into a Bourne-compatible syntax?

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -268,7 +268,6 @@ testpaths = "tests"
   cmd = """
     streamlit run
       --browser.gatherUsageStats false
-      --global.sharingMode off
```
lsorber, line 271: On closer inspection I do agree it is rather opinionated and would be OK to remove it. For my information, why do you prefer to remove this line?

lsorber (changes requested):
Thanks for this PR @tanguyNaets! I've added a few comments, most should be easy to resolve.

{{ cookiecutter.package_name|slugify }}/.gitignore
```diff
@@ -58,3 +58,6 @@ __pycache__/
 
 # VS Code
 .vscode/
+
+# Docs
+docs
```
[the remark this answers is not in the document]
tanguyNaets: Yes, I was thinking of avoiding all the html docs in the repo, since html is a bit ugly to track in git and it's not necessarily up to speed with the project if one forgets to build the docs before pushing and  these docs are generated from the code (docstrings, ...) which is the single source of truth one would want to maintain.
But I understand it's comfortable for external parties to view the docs directly from gitlab or github.
I can remove that change for now.
Would a separate PR enforcing the rebuild of the docs in the last pre-commit step make sense for you?

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
tanguyNaets: Yes

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: It's not recognized by streamlit anymore in the latest poetry compatible version of the project.
We have `streamlit = "^1.3.0"` in `.toml`, and streamlit 1.4 and 1.5, which are now available, and which are both compatible with the previous statement, do not have that flag anymore.

{{ cookiecutter.package_name|slugify }}/.gitignore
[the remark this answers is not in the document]
lsorber: Would remove this change for now indeed. As for a PR that builds the docs: I'm actually not fully satisfied yet with `pdoc`, perhaps we should standardise on another tool -- but let's discuss that offline.

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -73,6 +73,10 @@ COPY . .
 
 # Expose the application.
 {%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int %}
+
+RUN --mount=type=cache,target=/root/.cache \
+    pip install --no-input poethepoet
```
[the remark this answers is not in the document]
tanguyNaets: Unfortunately, `poe` is not available in the App image.
I tracked down where the issue starts to arise. In the ci layer, when we run `poetry install --no-dev --no-interaction`, it deletes `poe` but not the other binaries in  `/usr/local/bin` such as (pip and starship). It's because `poethepoet` is only a dev dependency, so poetry doesn't see it with `--no-dev` flag and decides to remove that package.
The simplest way to fix the issue is to move the poetry dependency from dev to main dependencies. It's done now and I removed the need to reinstall poethepoet

{{ cookiecutter.package_name|slugify }}/.gitignore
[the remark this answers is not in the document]
tanguyNaets: oki

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -32,6 +32,7 @@ version_files = ["pyproject.toml:version"]
 fastapi = "^0.70.1"
 gunicorn = "^20.1.0"
 {%- endif %}
+poethepoet = "^0.12.2"
```
lsorber, line 35: For non-deployable projects (i.e. projects that are Python packages), I'd prefer not to make `poethepoet` a required dependency. To solve that, could you wrap this line in:
```yaml
{%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int %}
poethepoet = "^0.12.2"
{%- endif %}
```
and then also add it as a conditional development dependency with:
```yaml
{%- if not cookiecutter.with_fastapi_api|int and not cookiecutter.with_streamlit_app|int %}
poethepoet = "^0.12.2"
{%- endif %}
```

{{ cookiecutter.package_name|slugify }}/Dockerfile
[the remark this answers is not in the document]
lsorber: I see, thanks for investigating! For posterity, this is the relevant Poetry issue: https://github.com/python-poetry/poetry/issues/3139.

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: Good idea. Done. Done :)

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -72,9 +75,11 @@ isort = "^5.10.1"
 jupyterlab = "^3.2.5"
 {%- endif %}
 mypy = "^0.931"
+{%- if not cookiecutter.with_fastapi_api|int and not cookiecutter.with_streamlit_app|int %}
+poethepoet = "^0.12.2"
```
lsorber, line 79: Could you put this one directly under `pep8-naming` so that we maintain the alphabetical ordering of dev dependencies? Thanks!

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
lsorber, line 62: 1. Could you remove the curly braces where they are not needed in this script (cf. example below)?
2. I saw you wrapped environment variables with `"`, which I think is a good idea -- would do that on this line too:
```suggestion
        for DOCKER_TAG in "$DOCKER_TAGS"; do
```

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: Done.

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
tanguyNaets: Done.

lsorber (approved):

PR #3, opened 2022-02-03
Tn ubuntu

tanguyNaets:
Replace debian base image by more secure ubuntu 20.04 image. (Should be merged after tn-small-fixes)

```bash
❯ trivy image ubuntu:20.04
ubuntu:20.04 (ubuntu 20.04)
===========================
Total: 69 (UNKNOWN: 0, LOW: 40, MEDIUM: 29, HIGH: 0, CRITICAL: 0)

❯ trivy image python:3.8-slim
python:3.8-slim (debian 11.2)
=============================
Total: 87 (UNKNOWN: 0, LOW: 12, MEDIUM: 27, HIGH: 35, CRITICAL: 13)
```


lsorber:
Hey @tanguyNaets, thanks for this PR! Before reviewing, I'd like to make sure that we're making the best choice for the base image. Based on this article that compares Python base images [1] and these two articles on Docker image vulnerability scanning [2] [3], my takeaway is that Python⇒Speed recommends using the official Python Docker images and upgrading the system dependencies.

[1] https://pythonspeed.com/articles/base-image-python-docker-images/
[2] https://pythonspeed.com/articles/docker-python-security-scan/
[3] https://pythonspeed.com/articles/docker-security-scanner/

tanguyNaets:
LSS: We cannot do much harm when going for any of those base images. It's more a matter of preference at that point. So I'm good keeping the official python base image if you are still ok with that and remove that PR.

Hi @lsorber , I've gone through your references. The main take-away is that we may want to test base image security with `trivy` using the `--ignore-unfixed` flag to avoid false positives, trivial cases, ....

Doing that, both ubuntu and official python images look pretty good.

```
❯ trivy image --light --exit-code 1 --ignore-unfixed python:3.8-slim
python:3.8-slim (debian 11.2)
=============================
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)


Python (python-pkg)
===================
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)



❯ trivy image --light --exit-code 1 --ignore-unfixed ubuntu:20.04
ubuntu:20.04 (ubuntu 20.04)
===========================
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```

The other concerns (having latest python available, the size and speed of the images, ...) are kind irrelevant when comparing the 3 main long term supported unix base images (ubuntu, debian and redhat) and the official python base image (based on debian as well) because they are all in the same ballpark.

Conclusion: We cannot do much harm when going for any of those base images. It's more a matter of preference at that point. So I'm good keeping the official python base image if you are still ok with that.

lsorber:
> So I'm good keeping the official python base image if you are still ok with that.
Yes, let's go with that given the analysis above. I'll go ahead and close this PR, but thanks for the initiative!

PR #4, opened 2022-02-08
Tn logs

tanguyNaets:
Add proper logs display for fastapi and streamlit applications. Especially useful when tracking bugs on deployed solutions.


{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py
```diff
@@ -1 +1,29 @@
...
+import coloredlogs
+
+
+def configure_root_logger() -> None:
+    """Configure the root logger."""
+    # Remove all handlers associated with the root logger object.
+    for handler in logging.root.handlers:
+        logging.root.removeHandler(handler)
+    # Add coloredlogs' coloured StreamHandler to the root logger.
+    coloredlogs.install(
```
lsorber, line 15: To simplify this function, I'd prefer to replace this call with `coloredlogs.install()` as I don't see anything in the configuration that we would really need to apply by default.

{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py
```diff
@@ -1 +1,29 @@
...
+            "name": {"color": "blue"},
+            "funcName": {"color": "magenta"},
+            "levelname": {"color": "black", "bold": True},
+        },
+        isatty=True,
+    )
+
+
+configure_root_logger()
+{%- endif %}
```
lsorber, line 29: Not sure if it's because of the way that GitHub renders files in the browser, but we need a newline after this `{%- endif %}` in order to satisfy the [end-of-file-fixer](https://github.com/pre-commit/pre-commit-hooks#end-of-file-fixer) linter.

lsorber (changes requested):
LGTM apart from a two minor comments.

{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py
tanguyNaets: Oki, changed to coloredlogs defaults.

{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py
tanguyNaets: Yes, its just the Github rendering. There is actually a new line :)

lsorber (approved):

PR #5, opened 2022-02-15
Use commit SHA provided by GitLab CI/CD

sinopeus:
GitLab CI/CD already has some variables available. This way the image hashes can map one-on-one to Git commit hashes (traceability, yay!).


sinopeus:
After discussion with @lsorber, the conclusion is that this change would trigger unnecessary extra Docker image builds, as the original logic only rebuilds the image when the 'base' layers of the image could change (as defined by the Dockerfile and Poetry project and lock files). This change would change the behavior to rebuilding on every change to the repository, including Python source code changes that shouldn't trigger a full rebuild. This is not necessary, as the test step handles that by copying the Python source code over into a container based on the pre-built Docker image (if it exists). Closing this PR 🙂

PR #6, opened 2022-02-17
Tn fix gitlab pipeline

tanguyNaets:
- Fix pipeline
In sh, a string variable is not split on the the spaces it contains in a for loop if wrapped in "". So `for SOMETHING in "$DOCKER_TAGS"`, doesn't perform a loop, even if `DOCKER_TAGS` contains spaces. Instead, if we want to split on spaces, we should use the following construct: `for SOMETHING in $DOCKER_TAGS`.
- Enforce the deploy stage of the gitlab ci pipeline to tag the deploy image with "latest", no matter the circumpstances. (Previous behavior: We do not have the tag: `latest` if we deploy a non-main branch before we merge to `main` and deploy again. It will only tag `latest` if we don't deploy a non-main branch, then merge and deploy the `main` branch.
- Put the app docker images in the `app` repository instead of in the root which gets named `Root image`


{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -127,8 +127,9 @@ Publish:
     - .docker
   stage: deploy
   variables:
+    DOCKER_PUSH: 1
     DOCKER_BASE_IMAGE: $CI_REGISTRY_IMAGE/ci:$CI_IMAGE_SHA
-    DOCKER_IMAGE: $CI_REGISTRY_IMAGE
+    DOCKER_IMAGE: $CI_REGISTRY_IMAGE/app
```
lsorber, line 132: I don't think this is necessary: the image may be the name of the repository itself (which I think is preferably because then the image name is that of the actual solution – otherwise with several projects you end up with a bunch of images all called `app`).

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -127,8 +127,9 @@ Publish:
     - .docker
   stage: deploy
   variables:
+    DOCKER_PUSH: 1
```
lsorber, line 130: Could you put this line in alphabetical order under the `variables:` section?

lsorber (changes requested):
Thanks for the PR! Two minor comments, and we're good to go.

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
tanguyNaets: That's what I thought as well, but it looks like gitlab needs to create a repository inside the registry of the project if we don't provide one. If we don't provide any, it will name it `Root image` and our repositories will look like that in our gitlab's project registry: 
image

Would you rather keep the default `Root image` repository, or name it `app` or something else which makes it easy to understand it's the application repository of the project?

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
tanguyNaets: Done

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
tanguyNaets: In practice `Root image` only shows in the gitlab UI. The repository is actually at the root of the registry. So I can remove it if you still think that makes more sense.

tanguyNaets: Oki, I removed it anyways. Should be good to merge :)

lsorber:
Merged, thanks @tanguyNaets!

PR #7, opened 2022-03-07
Tn fix darglint doctests

tanguyNaets:
- Ignore darglint `DAR103` because typing is already handled with mypy at the args level and there is no need to duplicate typing information in docstrings
- Enable docstests and have them included in the test coverage
- Hard-pin bandit because of breaking change when interacting with flake8
- Bump pdoc dependency to fix security breach


{{ cookiecutter.package_name|slugify }}/.flake8
```diff
@@ -7,5 +7,5 @@ max_line_length = 100
 max_complexity = 10
 docstring_style = numpy
 doctests = True
-ignore = E203,E501,FS003,S101,W503
+ignore = E203,E501,FS003,S101,W503,DAR103
```
lsorber, line 10: Could you put this one up front (so the list is alphabetically sorted)?
```suggestion
ignore = DAR103,E203,E501,FS003,S101,W503
```

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -58,10 +58,11 @@ uvicorn = { extras = ["standard"], version = "^0.16.0" }
 # https://python-poetry.org/docs/master/managing-dependencies/
 # TODO: Split in `tool.poetry.group.dev` and `tool.poetry.group.test` when Poetry 1.2.0 is released.
 [tool.poetry.dev-dependencies]
-cruft = "^2.10.1"
+bandit = "1.7.2"  # TODO: remove this line once this is solved: https://github.com/PyCQA/bandit/issues/841
```
lsorber, line 61: - I think this one is the original bandit issue (the one you link to appears to be a duplicate): https://github.com/PyCQA/bandit/issues/837. Doesn't look like bandit will solve this though since they don't consider the thing that is broken as part of their public API.
- And this is the flake8-bandit issue: https://github.com/tylerwince/flake8-bandit/issues/21. Would refer to this one instead, hoping that they can fix it there.
- Would also relax the bandit constraint a bit just in case an exact pin could cause the environment to be unsolvable:
```suggestion
bandit = "<1.7.3"  # TODO: Remove after https://github.com/tylerwince/flake8-bandit/issues/21
```

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -58,10 +58,11 @@ uvicorn = { extras = ["standard"], version = "^0.16.0" }
...
 black = "^22.1.0"
 commitizen = "^2.20.2"
 coverage = { extras = ["toml"], version = "^6.2" }
+cruft = "^2.10.1"
```
lsorber, line 65: Good catch

lsorber (changes requested):
Thanks for the improvements!

{{ cookiecutter.package_name|slugify }}/.flake8
tanguyNaets: Done

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: I updated the referenced issue.
Regarding the pinning, I can't set it to `"<1.7.3"` for now because both `1.7.3` and the latest version `1.7.4` still have the bug. But the hard pinning should not be an issue on the long run since the todo suggests to remove that dependency and trust flake8-bandit to install the right version in the future.

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: Thanks

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -58,10 +58,11 @@ uvicorn = { extras = ["standard"], version = "^0.16.0" }
 # https://python-poetry.org/docs/master/managing-dependencies/
 # TODO: Split in `tool.poetry.group.dev` and `tool.poetry.group.test` when Poetry 1.2.0 is released.
 [tool.poetry.dev-dependencies]
-cruft = "^2.10.1"
+bandit = "1.7.2"  # TODO: remove this line once this is solved: https://github.com/tylerwince/flake8-bandit/issues/21
```
lsorber, line 61: Apparently this has been fixed now, version 3.0.0 has been released 4 hours ago. Could you update this PR with that new version?

{{ cookiecutter.package_name|slugify }}/pyproject.toml
lsorber: Resolving this conversation, and opened a new one related to it.

{{ cookiecutter.package_name|slugify }}/pyproject.toml
tanguyNaets: Just tested the new version to be sure and in deed, it works :). Explicit bandit dependency removed and upgraded flake8-bandit dependency to 3.0.0.
Should be good to merge :)

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -58,14 +58,14 @@ uvicorn = { extras = ["standard"], version = "^0.16.0" }
...
 # TODO: Split in `tool.poetry.group.dev` and `tool.poetry.group.test` when Poetry 1.2.0 is released.
 [tool.poetry.dev-dependencies]
-cruft = "^2.10.1"
 black = "^22.1.0"
 commitizen = "^2.20.2"
 coverage = { extras = ["toml"], version = "^6.2" }
+cruft = "^2.10.1"
 darglint = "^1.8.1"
 flake8 = "^4.0.1"
 flake8-comprehensions = "^3.7.0"
```
lsorber, line 67: Could you move this one down after `flake8-bugbear`?

tanguyNaets: Done, moved around flake8-print as well while at it :p

PR #15, opened 2022-03-25
Improve SSH config instructions

lsorber:


{{ cookiecutter.package_name|slugify }}/README.md
```diff
@@ -18,7 +18,17 @@ To add and install this package as a dependency of your project, run `poetry add
 <summary>Setup: once per device</summary>
 
 {% if cookiecutter.continuous_integration == "GitLab" -%}
-1. [Generate an SSH key](https://docs.gitlab.com/ee/ssh/README.html#generate-an-ssh-key-pair) for GitLab, [add the SSH key to your GitLab account](https://docs.gitlab.com/ee/ssh/README.html#add-an-ssh-key-to-your-gitlab-account), and [add the SSH key to your authentication agent](https://docs.gitlab.com/ee/ssh/README.html#configure-ssh-to-point-to-a-different-directory).
+1. [Generate an SSH key](https://docs.gitlab.com/ee/ssh/README.html#generate-an-ssh-key-pair) and [add the SSH key to your GitLab account](https://docs.gitlab.com/ee/ssh/README.html#add-an-ssh-key-to-your-gitlab-account).
+1. Edit `~/.ssh/config` to configure SSH to automatically load your SSH keys:
```
tanguyNaets, line 22: Is there a need to reboot the mac, the shell or the ssh agent for the changes to take effect? I've read Raphael mentioned a reboot step.
If not, that's  good to merge for me :).
Otherwise, it might be useful to have a `2. Reboot your <ssh-agent|mac>` for the config changes to take effect.

tanguyNaets (commented):
Looks great @lsorber, thanks :)
Just a minor comment.

sinopeus (changes requested):
As mentioned in https://github.com/radix-ai/poetry-cookiecutter/pull/15/commits/0195a2be0e78efe36459e403ae8426a76b04a480#r836126133, probably best to remove the `Host gitlab.com` and `Host github.com` entries to allow maximum flexibility. Normally if any of your SSH keys can connect, you should be good to go with no need for extra config.

{{ cookiecutter.package_name|slugify }}/README.md
sinopeus: I think this is not necessary so no change needed here.

lsorber:
> As mentioned in [0195a2b#r836126133](https://github.com/radix-ai/poetry-cookiecutter/commit/0195a2be0e78efe36459e403ae8426a76b04a480#r836126133), probably best to remove the `Host gitlab.com` and `Host github.com` entries to allow maximum flexibility. Normally if any of your SSH keys can connect, you should be good to go with no need for extra config.

Interesting. To make sure I understand: will SSH try all your keys if you don’t specify which one to use? And if that’s the case, is there a downside of doing it that way?

{{ cookiecutter.package_name|slugify }}/README.md
lsorber: No reboot required, no (based on my own experience).

lsorber:
To answer my own question, [SSH will indeed just try all keys it finds](https://superuser.com/questions/268776/how-do-i-configure-ssh-so-it-doesnt-try-all-the-identity-files-automatically). The downside is that the default max number of authentication attempts is 6.

That means that we can choose to simplify the instructions, if we assume the user will have no more than 6 keys (give or take). You could make the argument that somebody with more than 7 keys will know to configure them properly, so it might actually be OK to simplify. Thoughts?

sinopeus:
Keep it simple - I'd just remove as it doesn't really add any value. And
let's indeed assume those with 7 SSH keys on the same machine have a pretty
good idea of what they're doing 😉

sinopeus (approved):
Just one last nitpick 😉 I would make the command a little bit more readable by using heredoc:

```
cat >> ~/.ssh/config << EOF
Host *
  AddKeysToAgent yes
  IgnoreUnknown UseKeychain
  UseKeychain yes
EOF
```

Apart from that, LGTM and go ahead with merging!

lsorber:
That's much better, thanks for the suggestion @sinopeus!

PR #17, opened 2022-03-25
Improve `coloredlogs` installation

lsorber:
- [x] Best practice: no code in `__init__.py` (except export definitions).
- [x] Best practice: don't run code on importing a package (now runs on API startup event).
- `coloredlogs` is no longer available by default for Streamlit, but the benefit there isn't as big as with FastAPI.


tanguyNaets (approved):
LGTM

sinopeus (approved):

PR #20, opened 2022-03-28
Refactor Sentry into a generic feature

lsorber:


{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/{% if cookiecutter.with_sentry_logging|int %}sentry.py{% endif %}
```diff
@@ -8,10 +8,9 @@ import sentry_sdk
 
 def configure_sentry() -> None:
     """Configure Sentry."""
-    environment = os.environ.get("ENVIRONMENT", "local")
     sentry_sdk.init(
-        dsn="{{ cookiecutter.sentry_dsn }}",
+        dsn=os.environ["SENTRY_DSN"],
```
tanguyNaets, line 12: Where does that variable come from?
Now that you don't ask "sentry_dsn" anymore at the creation of the project, the `dns` parameter here will always be `None` unless the developer exports a `SENTRY_DSN` environment variable (only available at `development` time but not at `app` time) right?

tanguyNaets (changes requested):

{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/{% if cookiecutter.with_sentry_logging|int %}sentry.py{% endif %}
sinopeus: I think the intention is that the deployer "feeds" this environment variable to the deployed Docker container. This way it's also not hardcoded, which is more flexible for deployment.

sinopeus (approved):

{{ cookiecutter.package_name|slugify }}/src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/{% if cookiecutter.with_sentry_logging|int %}sentry.py{% endif %}
tanguyNaets: Then shouldn't we have that documented in the readme for the deployer to know he/she has to set that variable in order to get sentry to work?

tanguyNaets: Oki, that's fair enough for me.

PR #22, opened 2022-03-29
fix:Force click version to max 8.0.4

sinopeus:


.github/workflows/test.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Cruft and Poetry
-        run: pip install --no-input cruft poetry
+        run: pip install --no-input cruft poetry "click<=8.0.4"  # pinned to work around https://github.com/tiangolo/typer/issues/377, change when that is fixed
```
lsorber, line 33: For my information, why is this needed here? Isn't it line 50 that will actually install click?

{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -48,6 +48,7 @@ streamlit = "^1.3.0"
 {%- endif %}
 {%- if cookiecutter.with_typer_cli|int %}
 typer = "^0.4.0"
+click = "<=8.0.4" # pinned to work around https://github.com/tiangolo/typer/issues/377, change when that is fixed
```
lsorber, line 51: Suggestion: let's give Typer until the end of today to fix this with a v0.4.1 before we merge this PR. (Then we don't need to rebase the other PRs).

lsorber (commented):
Thanks for the quick fix @sinopeus – one question and one suggestion before we merge.

.github/workflows/test.yml
sinopeus: They both install it. Check this run (commit no longer exists as I amended it, but it didn't have the change on this line): https://github.com/radix-ai/poetry-cookiecutter/runs/5735889706. If you check the "Install Cruft and Poetry", you can see that `click` gets installed in that step as well, with a version that has the bug.

{{ cookiecutter.package_name|slugify }}/pyproject.toml
sinopeus: That's a good idea. Let's hold off for today then!

lsorber: Update: will close this PR since Typer was updated today.

lsorber:
Closing since Typer was updated yesterday: https://twitter.com/tiangolo/status/1509269796222873604

PR #23, opened 2022-04-04
Install specific Poetry version

lsorber:
- [x] Use latest [Poetry installation instructions](https://github.com/python-poetry/poetry#installation).
- [x] Install specific version of Poetry to improve reproducibility and reduce the chance of incompatibilities when v1.2.0 is released.

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -25,10 +25,11 @@ ENV PYTHONUNBUFFERED 1
 WORKDIR /app/
 
 # Install base development environment with Poetry and Poe the Poet.
+ENV POETRY_VERSION 1.1.13
 ENV PATH /root/.local/bin/:$PATH
 RUN --mount=type=cache,target=/root/.cache \
     pip install --no-input --upgrade pip poethepoet && \
-    curl -sSL https://install.python-poetry.org | python - --version 1.1.13 && \
+    curl -sSL https://install.python-poetry.org | python - --version $POETRY_VERSION && \
     poetry config virtualenvs.create false && \
     poetry completions fish > ~/.config/fish/completions/poetry.fish && \
     poe _fish_completion > ~/.config/fish/completions/poe.fish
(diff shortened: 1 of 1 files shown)


sinopeus (approved):

sinopeus:
Maybe one more thing you could do is set an environment variable for the Poetry version (script reads out POETRY_VERSION from env vars). Then it's a bit more obvious where the Poetry version is being set.

lsorber:
> Maybe one more thing you could do is set an environment variable for the Poetry version (script reads out POETRY_VERSION from env vars). Then it's a bit more obvious where the Poetry version is being set.

Options I see:

(1) As an environment variable:
```Dockerfile
ENV POETRY_VERSION 1.1.13
```

(2) As a build arg:
```Dockerfile
ARG POETRY_VERSION
ENV POETRY_VERSION $POETRY_VERSION
```

(3) As a Cookiecutter setting.

I might be missing something, but I don't see a good reason to go for any of these 3 options:
- Compared to the current Dockerfile, (1) doesn't really provide any benefit that I can see.
- Making it a build arg as in (2) introduces complexity in the form of indirection, and it also doesn't seem to add anything we need (being able to dynamically set the Poetry version isn't something you're likely to do).
- Adding it to the Cookiecutter setting as in (3) does make it more explicit, but if we do that you cannot easily upgrade Poetry later on with `cruft update` because the value will be set in your `cookiecutter.json` settings.

Did you have something else in mind?

sinopeus:
I was thinking of option 1. I think putting it in an env var is just a tad bit more explicit and not hidden away at the end of a shell command. Functionally identical though, so either will definitely work!

lsorber:
@sinopeus Done!

sinopeus:
Great! And by the way, because of the env var, you don't have to specify the `--version` option to the installer script any more so that can go too 🙂

lsorber:
> Great! And by the way, because of the env var, you don't have to specify the `--version` option to the installer script any more so that can go too 🙂 And then it's time to merge this 😉

I'd rather be explicit about it, otherwise you'd have to know that setting the environment variable means that the installer will install that version.

PR #25, opened 2022-04-07
Improve Docker networking

### LINKED ISSUE #21: Automatically change port mapping if PORT 8000 is already used
When working on several poetry cookie cutter projects (and/or with other apps that use port 8000), the devcontainer won't work because the port 8000 is already used by those other projects/apps.

Simply manually setting a different port in the `docker-compose.yml` and `pyproject.toml` for each of the projects does the trick.

But it would be nice to have something like what jupyter notebook and lab do. I.e. trying their default port number, and using another one if the first one is already used, and another one if the second one is already in use, etc.
WDYT?

lsorber:
- [x] Add a hostname so you can easily access the service in the browser with the address `{package-slug}.local`.
- [x] Dynamically bind the container port to an available post on the host. Closes #21.
- [x] Expose the container's port to other containers with `expose`.
- [x] Adds `tty: true` for coloured output (default is false).

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/docker-compose.yml
+++ b/{{ cookiecutter.package_name|slugify }}/docker-compose.yml
@@ -12,6 +12,8 @@ services:
         {%- endif %}
       context: .
       target: dev
+    tty: true
+    entrypoint: []
     command:
       [
         "sh",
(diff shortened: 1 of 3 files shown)


{{ cookiecutter.package_name|slugify }}/docker-compose.yml
```diff
@@ -12,6 +12,8 @@ services:
         {%- endif %}
       context: .
       target: dev
+    tty: true
+    entrypoint: []
```
sinopeus, line 16: Alternatively: set the entrypoint like this `entrypoint: ["sh", "-c"]`, then you can simplify the `command` entry.

sinopeus (approved):

{{ cookiecutter.package_name|slugify }}/docker-compose.yml
lsorber: Good suggestion!

lsorber: Actually, on second thought it's not that easy: the entrypoint needs to be an executable, which means the command would still need to start with `["-c", "cp ..."]`, which is IMHO less clear than the way it is now.

PR #29, opened 2022-04-14
feat: use pre-commit's `default_install_hook_types`

lsorber:


{{ cookiecutter.package_name|slugify }}/pyproject.toml
```diff
@@ -81,7 +81,7 @@ pep8-naming = "^0.12.1"
 {%- if not cookiecutter.with_fastapi_api|int and not cookiecutter.with_streamlit_app|int %}
 poethepoet = "^0.12.2"
 {%- endif %}
-pre-commit = "^2.16.0"
+pre-commit = "^2.18.1"
```
sinopeus, line 84: You should probably also update `poetry.lock`!

lsorber: No need I think: `poetry.lock` is a generated file in this repo. It's only once you've scaffolded a project that `poetry.lock` is persisted.

sinopeus: I must have made my comment before my first cup of coffee in the morning, because there's indeed, obviously, no `poetry.lock` in this repo 🙂

PR #36, opened 2022-05-07
feat: use build and run time secrets

lsorber:
- [Docker Compose v2.5.0](https://github.com/docker/compose/releases/tag/v2.5.0) (released 2022-04-29) adds support for build time secrets.
- [Docker Desktop 4.8.0](https://docs.docker.com/desktop/mac/release-notes/#docker-desktop-480) (released 2022-05-06) ships with Docker Compose v2.5.0.
- Before this PR, build time secrets were implemented for the Docker CLI only. For Docker Compose, build time secrets were implemented as environment variables.
- Before this PR, run time secrets were implemented as environment variables.
- After this PR, build and run time secrets are supported for both the Docker CLI as well as Docker Compose.
- This PR also simplifies the one-time setup in a few ways:
  - Poetry credentials are now managed in Poetry's `auth.toml`.
  - It is no longer needed to configure Docker & Docker Compose to use BuildKit on macOS and Windows since this it is enabled by default in Docker Desktop on those platforms.

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -29,18 +29,6 @@ RUN --mount=type=cache,target=/root/.cache/ \
     curl -sSL https://install.python-poetry.org | python - --version $POETRY_VERSION && \
     poetry config virtualenvs.create false
 
-# Enable Poetry to publish to {% if cookiecutter.private_package_repository_name %}and install from a private package repository{% else %}PyPI{% endif %} [1].
-# [1] https://pythonspeed.com/articles/build-secrets-docker-compose/
-{%- if cookiecutter.private_package_repository_name %}
-ARG POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_USERNAME
-ENV POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_USERNAME $POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_USERNAME
-ARG POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD
-ENV POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD $POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD
-{%- else %}
-ARG POETRY_PYPI_TOKEN_PYPI
-ENV POETRY_PYPI_TOKEN_PYPI $POETRY_PYPI_TOKEN_PYPI
-{%- endif %}
-
 # Let Poe the Poet know it doesn't need to activate the Python environment.
 ENV POETRY_ACTIVE 1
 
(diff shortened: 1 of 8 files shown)


{{ cookiecutter.package_name|slugify }}/docker-compose.yml
```diff
@@ -3,24 +3,28 @@ version: "3.9"
 services:
   dev:
     build:
-      args:
```
sinopeus, line 6: 🎉

{{ cookiecutter.package_name|slugify }}/README.md
```diff
@@ -56,37 +62,29 @@ To add and install this package as a dependency of your project, run `poetry add
...
-
-    # fish
-    echo "set --export DOCKER_BUILDKIT 1" >> ~/.config/fish/config.fish
-    echo "set --export COMPOSE_DOCKER_CLI_BUILD 1" >> ~/.config/fish/config.fish
-    
-    # zsh
-    echo "export DOCKER_BUILDKIT=1" >> ~/.zshenv
-    echo "export COMPOSE_DOCKER_CLI_BUILD=1" >> ~/.zshenv
-    ```
+1. [Install Docker Desktop](https://www.docker.com/get-started). If you are using Linux, you must [configure Docker and Docker Compose to use the BuildKit build system](https://pythonspeed.com/articles/docker-buildkit/). On macOS and Windows, BuildKit is enabled by default in Docker Desktop.
```
sinopeus, line 87: From which version on is it default in Docker Desktop?

{{ cookiecutter.package_name|slugify }}/README.md
```diff
@@ -56,37 +62,29 @@ To add and install this package as a dependency of your project, run `poetry add
...
-    echo "export POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD='{password}'" >> ~/.bash_profile
-    
-    # fish
-    echo "set --export POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_USERNAME '{username}'" >> ~/.config/fish/config.fish
-    echo "set --export POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD '{password}'" >> ~/.config/fish/config.fish
-
-    # zsh
-    echo "export POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_USERNAME='{username}'" >> ~/.zshenv
-    echo "export POETRY_HTTP_BASIC_{{ cookiecutter.private_package_repository_name|upper|replace("-", "_") }}_PASSWORD='{password}'" >> ~/.zshenv
+    # Linux
```
sinopeus, line 76: The Cookiecutter now assumes that we have Poetry installed on our machines as well. Well, kind of, it's just a config file after all, but it is in a very specific location! I suppose we should make that explicit?

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -29,18 +29,6 @@ RUN --mount=type=cache,target=/root/.cache/ \
     curl -sSL https://install.python-poetry.org | python - --version $POETRY_VERSION && \
     poetry config virtualenvs.create false
 
-# Enable Poetry to publish to {% if cookiecutter.private_package_repository_name %}and install from a private package repository{% else %}PyPI{% endif %} [1].
```
sinopeus, line 32: Ahhhh, clean ...

{{ cookiecutter.package_name|slugify }}/README.md
lsorber: From the article:
image

{{ cookiecutter.package_name|slugify }}/README.md
lsorber: Indeed, it's not required to have Poetry installed on the host system, but the way of storing the credentials is indeed Poetry-specific. By storing the credentials in that location, we have the benefit that if you _do_ happen to use Poetry on the host system, that authentication will work out of the box on the host as well.

What would you like to make more explicit exactly?

lsorber:
@sinopeus I test-drove this PR with
```
cruft create -f git@github.com:radix-ai/poetry-cookiecutter --checkout ls-secrets
```
and then added a few improvements:
- Fixed the `auth.toml` location in the `.env` file (path needs to be quoted because of whitespace in the path).
- Fixed the image not building if the project is scaffolded without a git repo.
- Added an instruction to make sure to enable Docker Compose V2 in the Docker Desktop settings. This is necessary because VS Code still uses `docker-compose` for its commands (which is Docker Compose V1) instead of `docker compose` (which is Docker Compose V2). If you enable Docker Compose V2, then Docker will alias `docker-compose` to `docker compose`.

sinopeus (approved):

PR #46, opened 2022-05-24
fix: minimize devcontainer settings

lsorber:
This PR minimises the `devcontainer.json` and by doing so also fixes the startup warning (specifically, by not explicitly setting the language server to `Pylance` – which is the default anyway):
> The Pylance extension is not installed but the python.languageServer value is set to "Pylance". Would you like to install the Pylance extension to use Pylance, or revert back to Jedi?


lsorber:
Note: the build is failing because Starship's install script is broken, because they didn't upload binaries along with their 1.7.0 release.

sinopeus (approved):

PR #55, opened 2022-06-08
Tn fix secrets

tanguyNaets:


{{ cookiecutter.package_name|slugify }}/docker-compose.yml
```diff
@@ -5,10 +5,6 @@ services:
     build:
       context: .
       target: dev
-      {%- if cookiecutter.private_package_repository_name %}
-      secrets:
-        - poetry_auth
-      {%- endif %}
```
lsorber, line 11: I think we need to keep these: these are build-time secrets that are used to access private packages when building the Docker image (a feature that was [recently introduced with Docker Compose v2.5.0](https://github.com/radix-ai/poetry-cookiecutter/pull/36)).

{{ cookiecutter.package_name|slugify }}/docker-compose.yml
```diff
@@ -43,15 +39,15 @@ services:
     build:
       context: .
       target: app
-      {%- if cookiecutter.private_package_repository_name %}
-      secrets:
-        - poetry_auth
-      {%- endif %}
```
lsorber, line 49: Same comment: I think we need these build-time secrets here so that we can install private packages when building the app image.

{{ cookiecutter.package_name|slugify }}/docker-compose.yml
```diff
@@ -43,15 +39,15 @@ services:
...
     tty: true
     {%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int %}
     ports:
       - "8000:8000"
     {%- endif %}
+    {%- if cookiecutter.private_package_repository_name %}
+    secrets:
+      - poetry_auth
+    {%- endif %}
```
lsorber, line 50: These are run-time secrets, which I don't think are necessary for the app image since you normally won't need to install additional dependencies in the app image at run-time. But let me know if I'm missing something!

lsorber (commented):
Thanks for the PR @tanguyNaets! From the changes it's not clear what the PR's goal is, could you add a brief summary?

sinopeus:
In summary:

- We shouldn't remove the `poetry_auth` secret at build time or the build steps with `poetry install` will fail (when using a private package repository).
- There is (in principle) no need to add the `poetry_auth` secret at run time within the `app` profile, as there's no need to run `poetry install` with that profile.

@tanguyNaets Since both changes seem to be unneeded, is it OK if I close this PR? Do let us know if there's a use case for adding the `poetry_auth` secret when using the `app` profile!

sinopeus:
OK, thanks for the update!

PR #57, opened 2022-06-10
feat: better dev containers

lsorber:
Improvements:
1. [x] You can now run `docker compose run --rm dev` to start a Dev Container from the terminal. This is useful if you want to work from within the Dev Container without using VS Code or PyCharm. Also, if you use PyCharm, you will probably need to run the Dev Container alongside it like this to actually run commands like `poetry add` or `git commit` from within the environment.
    - `stdin_open: true` is added to the `dev` service to enable (1) to drop you into stdin (equivalent to `docker run -i`).
    - `zsh` is now the default command in `docker-compose.yml` so that you drop into `zsh` with `docker compose run`.
2. [x] The host's SSH agent is mounted again, otherwise a Terminal-based Dev Container (1) would not be very useful.
3. [x] Because (1) interferes with VS Code's Dev Containers, we now add `"overrideCommand": true` to `devcontainer.json`. This has the effect of starting the container with `sleep infinity` so that it doesn't quit. Because this replaces the startup command, we also need a `postStartCommand` to copy the files we need into the mounted volume as we do in the `docker-compose.yml`.
4. [x] `zsh -c 'source ~/.zshrc'` is added to the `Dockerfile` to speed up starting the Dev Container, since all zsh plugins will have been installed at build time.
5. [x] The `README.md` has been improved with 3 options to run your Dev Container: VS Code, PyCharm, and Terminal.
6. [x] The `dev` service now has a `dev` profile, and the `app` service no longer has any profile. This allows you to run `docker compose up` without any arguments and it will start the `app` service only (because it will not start any services with other profiles unless you ask it to).

Replaces #56.

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/.devcontainer/devcontainer.json
+++ b/{{ cookiecutter.package_name|slugify }}/.devcontainer/devcontainer.json
@@ -4,6 +4,8 @@
     "service": "dev",
     "workspaceFolder": "/app/",
     "initializeCommand": "ssh-add --apple-load-keychain || true",
+    "overrideCommand": true,
+    "postStartCommand": "cp --update /var/lib/poetry/poetry.lock /app/ && mkdir -p /app/.git/hooks/ && cp --update /var/lib/git/* /app/.git/hooks/{% if cookiecutter.private_package_repository_name %} && ln -s /run/secrets/poetry_auth /root/.config/pypoetry/auth.toml{% endif %}",
     "extensions": [
         "bungcip.better-toml",
         "eamodio.gitlens",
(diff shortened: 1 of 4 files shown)


sinopeus (approved):
So I I understand it correctly, VS Code will now use what used to be the `app` profile, which doesn't have the SSH and gitconfig mounts, since that's handled by VS Code itself. And there is now an explicit `dev` profile which is useful for directly opening a terminal in the container. Didn't know about `overrideCommand`! LGTM

lsorber:
> So I I understand it correctly, VS Code will now use what used to be the `app` profile, which doesn't have the SSH and gitconfig mounts, since that's handled by VS Code itself. And there is now an explicit `dev` profile which is useful for directly opening a terminal in the container. Didn't know about `overrideCommand`! LGTM

This was actually a mistake: VS Code apparently runs all the services in `docker-compose.yml`, except those with a `profile`. That resulted in running the `app` service instead of the `dev` service. I hotfixed it by undoing the `profile` change with this PR: https://github.com/radix-ai/poetry-cookiecutter/pull/58.

PR #60, opened 2022-06-17
feat: Improve Docker caching on GitLab CI

sinopeus:
- Introduce a `deps` stage: contains Poetry dependencies
- Change the `ci` stage to also include dev dependencies (later only test dependencies)
- Completely skip the build stage if no changes have been made to Docker or Poetry files using `rules:changes`
- Use `--cache-from` + CI image SHA to speed up the `test` and `app` Docker build (`test` gets much faster because we use a cached version of the `dev` stage)

This approach has given me some really nice speedup on GitLab CI:

- We don't even start the build stage if the relevant files haven't changed, and therefore there's no need to spin up `docker-in-docker`, which saves time. This did require moving the hash computation out into a separate stage.
- The `ci` stage contains dev dependencies as well, so `poetry install` in the `test` stage is instant.
- The `app` stage is not based on the matching `ci:<SHA1 hash>` image, but is rebuilt using the `--cache-from`; by doing this we still get a build speedup as the stages up to `deps` are shared with the `ci` image and we can use its corresponding cached layers.

I've tested it on GitLab CI manually and it looks good, but an automated check might be better - not sure how we would implement that in an elegant way.

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -23,7 +23,7 @@ FROM base as deps
 
 # Install the run time Python environment.
 # TODO: Replace `--no-dev` with `--without test` when Poetry 1.2.0 is released.
-COPY poetry.lock pyproject.toml /app/
+COPY poetry.lock* pyproject.toml /app/
 RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
     {%- if cookiecutter.private_package_repository_name %}
     --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -59,6 +59,7 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
     echo 'eval "$(starship init zsh)"' >> ~/.zshrc && \
     zsh -c 'source ~/.zshrc'
 
+# Persist output generated during docker build so that we can restore it in the dev container.
 RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
     git init && pre-commit install --install-hooks && \
     mkdir -p /var/lib/git/ && cp .git/hooks/commit-msg .git/hooks/pre-commit /var/lib/git/
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -38,7 +38,6 @@ RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
 FROM deps as ci
 
 # Install the development Python environment.
-COPY .pre-commit-config.yaml /app/
 RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
     {%- if cookiecutter.private_package_repository_name %}
     --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
@@ -60,6 +59,8 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
     echo 'antigen apply' >> ~/.zshrc && \
     echo 'eval "$(starship init zsh)"' >> ~/.zshrc && \
     zsh -c 'source ~/.zshrc'
+
+COPY .pre-commit-config.yaml /app/
 
 # Persist output generated during docker build so that we can restore it in the dev container.
 RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -29,7 +29,6 @@ RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
     {%- endif %}
     mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
     poetry install --no-dev --no-interaction
-{%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
 
 FROM deps as ci
 
@@ -63,6 +62,7 @@ RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
     git init && pre-commit install --install-hooks && \
     mkdir -p /var/lib/git/ && cp .git/hooks/commit-msg .git/hooks/pre-commit /var/lib/git/
 
+{%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
 FROM deps AS app
 
 # Copy the package source code to the working directory.
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -38,7 +38,6 @@ RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
     {%- if cookiecutter.private_package_repository_name %}
     --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
     {%- endif %}
-    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
     poetry install --no-interaction
 
 FROM ci as dev
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -66,8 +66,8 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
 RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
     git init && pre-commit install --install-hooks && \
     mkdir -p /var/lib/git/ && cp .git/hooks/commit-msg .git/hooks/pre-commit /var/lib/git/
-
 {%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
+
 FROM deps AS app
 
 # Copy the package source code to the working directory.
--- a/{{ cookiecutter.package_name|slugify }}/Dockerfile
+++ b/{{ cookiecutter.package_name|slugify }}/Dockerfile
@@ -19,7 +19,31 @@ ENV VIRTUAL_ENV /opt/app-env
 # Set the working directory.
 WORKDIR /app/
 
-FROM base as dev
+FROM base as deps
+
+# Install the run time Python environment.
+# TODO: Replace `--no-dev` with `--without test` when Poetry 1.2.0 is released.
+COPY poetry.lock pyproject.toml /app/
+RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
+    {%- if cookiecutter.private_package_repository_name %}
+    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
+    {%- endif %}
+    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
+    poetry install --no-dev --no-interaction
+{%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
+
+FROM deps as ci
+
+# Install the development Python environment.
+COPY .pre-commit-config.yaml poetry.lock* pyproject.toml /app/
+RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
+    {%- if cookiecutter.private_package_repository_name %}
+    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
+    {%- endif %}
+    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
+    poetry install --no-interaction
+
+FROM ci as dev
 
 # Install development tools: compilers, curl, git, gpg, ssh, starship, vim, and zsh.
 RUN rm /etc/apt/apt.conf.d/docker-clean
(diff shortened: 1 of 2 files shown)


sinopeus:
This approach has given me some really nice speedup on GitLab CI - no need to spin up `docker-in-docker` + `poetry install` in `test` stage is instant. Not 100% happy with the names of the Docker build stages now but wanted to get it out there before I go on holiday 🙂

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -13,48 +22,49 @@ stages:
...
+      done
+
+      # Build the Docker image with all of the selected tags.
+      {%- if cookiecutter.private_package_repository_name %}
+      echo "[http-basic.{{ cookiecutter.private_package_repository_name|slugify }}]" >> auth.toml
+      echo "username = \"gitlab-ci-token\"" >> auth.toml
+      echo "password = \"$CI_JOB_TOKEN\"" >> auth.toml
+      {%- endif %}
+      DOCKER_BUILDKIT=1 docker build \
+        --cache-from "${DOCKER_BASE_IMAGE:-ci}" \
```
lsorber, line 47: - Should `ci` be `deps` here?
- Does this actually speed anything up if the value is `ci` or `deps`?
- When/where would we set `DOCKER_BASE_IMAGE`?

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -13,48 +22,49 @@ stages:
     DOCKER_REGISTRY_USER: $CI_REGISTRY_USER
     DOCKER_REGISTRY_PASSWORD: $CI_REGISTRY_PASSWORD
   script:
-    - DOCKER_IMAGE_SHA="${DOCKER_IMAGE_SHA:-$(sha1sum Dockerfile poetry.lock pyproject.toml | sha1sum | cut -c 1-12)}"
-    - echo "CI_IMAGE_SHA=$DOCKER_IMAGE_SHA" >> .env
     - |
       echo "$DOCKER_REGISTRY_PASSWORD" | docker login --username "$DOCKER_REGISTRY_USER" --password-stdin "$DOCKER_REGISTRY"
       DOCKER_PUSH=${DOCKER_PUSH:-$(timeout 2s docker pull "$DOCKER_IMAGE":"$DOCKER_IMAGE_SHA" >/dev/null 2>&1 && echo $? || echo $?)}
-      if [ "$DOCKER_PUSH" -ne 1 ]; then
-        echo "$DOCKER_IMAGE:$DOCKER_IMAGE_SHA exists, skipping this job..."
```
lsorber, line 22: For my information: why do we no longer need/want an early exit if the image already exists?

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -13,48 +22,49 @@ stages:
...
-        for DOCKER_TAG in $DOCKER_TAGS; do
-          docker push "$DOCKER_IMAGE:$DOCKER_TAG"
-        done
-      fi
+      # Push all the tagged images.
+      for DOCKER_TAG in $DOCKER_TAGS; do
+        docker push "$DOCKER_IMAGE:$DOCKER_TAG"
+      done
+  rules:
+    - changes:
```
lsorber, line 64: Do I understand correctly that only changes to the files listed below _within the active branch_ will trigger this job? What happens on the first commit of a new branch?

lsorber (commented):
Thanks for the PR @sinopeus! I was actually thinking of something similar over the weekend, but you implemented it better I think.

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
lsorber: This article shows that of we want to benefit from the Docker cache, we need to build and push all stages: https://pythonspeed.com/articles/faster-multi-stage-builds/

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -19,7 +19,31 @@ ENV VIRTUAL_ENV /opt/app-env
...
+# Install the run time Python environment.
+# TODO: Replace `--no-dev` with `--without test` when Poetry 1.2.0 is released.
+COPY poetry.lock pyproject.toml /app/
+RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
+    {%- if cookiecutter.private_package_repository_name %}
+    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
+    {%- endif %}
+    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
+    poetry install --no-dev --no-interaction
+{%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
```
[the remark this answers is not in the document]
sinopeus: Must have slipped by!

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
sinopeus: Here's the docs:

https://docs.gitlab.com/ee/ci/yaml/#ruleschanges

The `rules:changes` conditions will be checked every time a push happens. First commit of a new branch should therefore not rerun the Docker build. However, the docs indicate that things like scheduled runs, manual runs, tag builds, ... will ignore `rules:changes`. I'll re-add the check for an existing `ci` image, but leaving the `rules:changes` in will still give large time savings (since the job doesn't even need to be executed).

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
sinopeus: The idea was that since the `rules:changes` (https://github.com/radix-ai/poetry-cookiecutter/pull/60#discussion_r902332721) block avoids rebuilds if there are no changes to the relevant files for the Docker build, it wasn't necessary to check the cache anymore. However since builds that weren't triggered by a push don't look at `rules:changes`, that doesn't always fly. I'll re-add it.

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
sinopeus: I think that blog post applies to multi-stage builds where a stage is build-only and "discarded" (its layers are not included in the final image). But for this Dockerfile the layers are "sequential" - `deps` contains `base`, `ci` and `app` contains `deps`, `dev` contains `ci`.

sinopeus: I fixed the issue with DOCKER_BASE_IMAGE - that was indeed not set (by accident), and no speedup would result (`ci` or `deps` weren't cached of course).

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -19,7 +19,28 @@ ENV VIRTUAL_ENV /opt/app-env
...
+    {%- if cookiecutter.private_package_repository_name %}
+    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
+    {%- endif %}
+    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
+    poetry install --no-dev --no-interaction
+
+FROM deps as ci
+
+# Install the development Python environment.
+COPY .pre-commit-config.yaml /app/
```
lsorber, line 36: I don't think this line is necessary in the `ci` stage?

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -35,32 +56,13 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
...
-COPY .pre-commit-config.yaml poetry.lock* pyproject.toml /app/
-RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
-    {%- if cookiecutter.private_package_repository_name %}
-    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
-    {%- endif %}
-    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
-    poetry install --no-interaction && \
-    mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
+# Persist output generated during docker build so that we can restore it in the dev container.
+RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
```
lsorber, line 60: The `.pre-commit-config.yaml` should be copied here, because we'll use it to install pre-commit on line 61.

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -35,32 +56,13 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
...
-# Install the run time Python environment.
-# TODO: Replace `--no-dev` with `--without test` when Poetry 1.2.0 is released.
-COPY poetry.lock pyproject.toml /app/
-RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
-    {%- if cookiecutter.private_package_repository_name %}
-    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
-    {%- endif %}
-    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
-    poetry install --no-dev --no-interaction
 {%- if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int or cookiecutter.with_typer_cli|int %}
```
lsorber, line 64: Nit-pick, but the blank line above this line should be removed and a blank line below this line should be added so that the two stages don't 'stick' to each other when rendered by cookiecutter.

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -13,11 +22,10 @@ stages:
     DOCKER_REGISTRY_USER: $CI_REGISTRY_USER
     DOCKER_REGISTRY_PASSWORD: $CI_REGISTRY_PASSWORD
   script:
-    - DOCKER_IMAGE_SHA="${DOCKER_IMAGE_SHA:-$(sha1sum Dockerfile poetry.lock pyproject.toml | sha1sum | cut -c 1-12)}"
-    - echo "CI_IMAGE_SHA=$DOCKER_IMAGE_SHA" >> .env
     - |
       echo "$DOCKER_REGISTRY_PASSWORD" | docker login --username "$DOCKER_REGISTRY_USER" --password-stdin "$DOCKER_REGISTRY"
       DOCKER_PUSH=${DOCKER_PUSH:-$(timeout 2s docker pull "$DOCKER_IMAGE":"$DOCKER_IMAGE_SHA" >/dev/null 2>&1 && echo $? || echo $?)}
+
```
lsorber, line 28:
```suggestion
```

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -55,6 +64,11 @@ stages:
           docker push "$DOCKER_IMAGE:$DOCKER_TAG"
         done
       fi
+  rules:
+    - changes:
+        - Dockerfile
+        - pyproject.toml
+        - poetry.lock
```
lsorber, line 71: Nit-pick, but would sort the list alphabetically

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
lsorber: OK, got it, thanks!

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -19,7 +19,28 @@ ENV VIRTUAL_ENV /opt/app-env
...
+FROM base as deps
+
+# Install the run time Python environment.
+COPY poetry.lock* pyproject.toml /app/
+RUN --mount=type=cache,id=poetry,target=/root/.cache/ \
+    {%- if cookiecutter.private_package_repository_name %}
+    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
+    {%- endif %}
+    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
+    poetry install --no-dev --no-interaction
```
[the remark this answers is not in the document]
sinopeus: Added

{{ cookiecutter.package_name|slugify }}/Dockerfile
sinopeus: I think we need it as we run `poe lint` using that image, which runs `pre-commit`.

{{ cookiecutter.package_name|slugify }}/Dockerfile
sinopeus: Fixed

{{ cookiecutter.package_name|slugify }}/Dockerfile
lsorber: Yes, but shouldn't it use the PR branch's `.pre-commit-config.yaml`, not the one that happens to be in the image?

{{ cookiecutter.package_name|slugify }}/Dockerfile
lsorber: Replied to the original comment

{{ cookiecutter.package_name|slugify }}/Dockerfile
sinopeus: Correct. Fixed

{{ cookiecutter.package_name|slugify }}/Dockerfile
sinopeus: Resolved original comment 🙂

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -35,32 +60,15 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
...
-    {%- if cookiecutter.private_package_repository_name %}
-    --mount=type=secret,id=poetry_auth,target=/root/.config/pypoetry/auth.toml \
-    {%- endif %}
-    mkdir -p src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/ && touch src/{{ cookiecutter.package_name|slugify|replace("-", "_") }}/__init__.py && touch README.md && \
-    poetry install --no-interaction && \
-    mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
+COPY .pre-commit-config.yaml /app/
+
+# Persist output generated during docker build so that we can restore it in the dev container.
+RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
```
lsorber, line 66: Suggest to colocate the `COPY` statment as part of the `# Persist output` block:
```suggestion
# Persist output generated during docker build so that we can restore it in the dev container.
COPY .pre-commit-config.yaml /app/
RUN mkdir -p /var/lib/poetry/ && cp poetry.lock /var/lib/poetry/ && \
```

{{ cookiecutter.package_name|slugify }}/Dockerfile
```diff
@@ -19,7 +19,32 @@ ENV VIRTUAL_ENV /opt/app-env
 # Set the working directory.
 WORKDIR /app/
 
-FROM base as dev
+FROM base as deps
+
+RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
```
lsorber, line 24: Suggest to add a comment to explain this block:
```suggestion

# Install compilers that may be required for certain packages or platforms.
RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt/ \
```

{{ cookiecutter.package_name|slugify }}/{% if cookiecutter.continuous_integration == "GitLab" %}.gitlab-ci.yml{% endif %}
```diff
@@ -39,6 +46,7 @@ stages:
         echo "password = \"$CI_JOB_TOKEN\"" >> auth.toml
         {%- endif %}
         DOCKER_BUILDKIT=1 docker build \
+          --cache-from "$CI_REGISTRY_IMAGE/ci:$CI_IMAGE_SHA" \
```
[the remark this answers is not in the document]
lsorber: That line only checks if the image is available and times out after at most 2 seconds, so it won't cover this unfortunately. But the Dev Container CLI will come to our rescue in a future PR :-).

lsorber (changes requested):
Final nitpicks and we're good to go!

lsorber (approved):
Approved!

PR #63, opened 2022-06-18
feat: Run tests in Docker

sinopeus:
- Build Docker CI image
- Run tests in built Docker CI image
- Use GitHub Actions Docker cache to speed up future Docker builds
- Drive-by action upgrades :-)

### DIFF
--- a/{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
+++ b/{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
@@ -9,48 +9,46 @@ on:
 
 jobs:
 
-  test:
-
+  build:
     runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build dev Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          target: dev
+          push: true
+          cache-from: type=gha  
+          cache-to: type=gha,mode=max
+          tags: |
+            ghcr.io/${{ github.repository }}/ci:${{ github.sha }}
 
-    strategy:
-
-      matrix:
-        python-version: ["{{ cookiecutter.python_version }}"]
-      fail-fast: false
+  test:
+    needs: build
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/${{ github.repository }}/ci:${{ github.sha }}
 
     steps:
 
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up Python {% raw %}${{ matrix.python-version }}{% endraw %}
-        uses: actions/setup-python@v3
-        with:
-          python-version: {% raw %}${{ matrix.python-version }}{% endraw %}
-
-      - name: Cache Python environment
-        uses: actions/cache@v3
-        with:
-          path: .venv/
-          key: {% raw %}venv-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('poetry.lock') }}{% endraw %}
-
-      - name: Install Poetry
-        run: pip install --no-input poetry
-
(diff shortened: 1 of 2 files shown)


sinopeus:
Still need to fix private package repo support + other things that are a bit broken. But that will be for another time, just wanted to put this out there 🙂

lsorber:
> Still need to fix private package repo support + other things that are a bit broken. But that will be for another time, just wanted to put this out there 🙂

OK, let me know when it's ready for review!

{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
```diff
@@ -9,48 +9,46 @@ on:
...
-
     steps:
-
       - name: Checkout
         uses: actions/checkout@v3
-
-      - name: Set up Python {% raw %}${{ matrix.python-version }}{% endraw %}
-        uses: actions/setup-python@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
```
lsorber, line 18: For my information: why is this action needed? I believe `docker/build-push-action` already uses BuildKit, if that's the reason.

{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
```diff
@@ -9,48 +9,46 @@ on:
...
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build dev Docker image
+        uses: docker/build-push-action@v2
         with:
-          path: .venv/
-          key: {% raw %}venv-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('poetry.lock') }}{% endraw %}
+          context: .
+          target: dev
```
lsorber, line 29: Shouldn't we be building `ci` here? The `dev` image is much larger because it contains developer tools that in principle aren't necessary in CI.

{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
```diff
@@ -9,48 +9,46 @@ on:
...
 
-  test:
-
+  build:
     runs-on: ubuntu-latest
-
-    strategy:
-
-      matrix:
-        python-version: ["{{ cookiecutter.python_version }}"]
```
lsorber, line 19: Although I'm all for using Docker, we do have to be aware that this PR removes the ability to test in a matrix, which is a bit of a downgrade for Python packages (for Python applications that shouldn't matter too much).

{{ cookiecutter.package_name|slugify }}/.github/workflows/test.yml
```diff
@@ -9,48 +9,46 @@ on:
...
         with:
-          path: .venv/
-          key: {% raw %}venv-${{ runner.os }}-py${{ matrix.python-version }}-${{ hashFiles('poetry.lock') }}{% endraw %}
+          context: .
+          target: dev
+          push: true
+          cache-from: type=gha  
+          cache-to: type=gha,mode=max
+          tags: |
+            ghcr.io/${{ github.repository }}/ci:${{ github.sha }}
```
lsorber, line 34: What is this SHA based on? Shouldn't we base it off of `Dockerfile`, `poetry.lock`, and `pyproject.toml` as we do in GitLab CI/CD?

{{ cookiecutter.package_name|slugify }}/.github/workflows/{% if cookiecutter.with_fastapi_api|int or cookiecutter.with_streamlit_app|int %}deploy.yml{% endif %}
```diff
@@ -38,29 +38,27 @@ jobs:
       - name: Log in to the Docker registry
         uses: docker/login-action@v1
         with:
-          registry: {% raw %}${{ env.REGISTRY }}{% endraw %}
-          username: {% raw %}${{ github.actor }}{% endraw %}
-          password: {% raw %}${{ secrets.GITHUB_TOKEN }}{% endraw %}
```
lsorber, line 43: We need the `{% raw %}` bits here because otherwise Cookiecutter will treat these as values that should be interpolated.

lsorber (changes requested):
Thanks for the PR! I know it's not ready yet, but I already did a first review pass.

PR #65, opened 2022-06-27
ci: test with devcontainer CLI

lsorber:
Uses the new [devcontainers/ci](https://github.com/devcontainers/ci) GitHub Action to test the Dev Container directly, instead of creating a local Python environment as we do now.

EDIT: Because of https://github.com/devcontainers/ci/issues/148, we'll use the devcontainers CLI instead.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,71 +7,62 @@ on:
   pull_request:
 
 jobs:
-
   test:
-
     runs-on: ubuntu-latest
 
     strategy:
-
-      matrix:
-        python-version: ["3.8"]
       fail-fast: false
+      matrix:
+        python-version: ["3.8", "3.10"]
+
+    name: Python ${{ matrix.python-version }}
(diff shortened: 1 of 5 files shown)


lsorber:
Update: this PR should be good to go, but the CI will fail until [requests v2.28.1 is released](https://github.com/psf/requests/pull/6182).

sinopeus (approved):
Oh, that's easy to use! Looks good to me. Do the underlying Docker builds use cache? 🙂 In any case it's already a very nice improvement so let's get this merged.

lsorber:
> Oh, that's easy to use! Looks good to me. Do the underlying Docker builds use cache? 🙂 In any case it's already a very nice improvement so let's get this merged.

Thanks! No, I don't think this uses any Docker image caching yet. Once https://github.com/devcontainers/ci/issues/148 is resolved, we can switch to the GitHub Action, which should benefit from image caching.

Issue #78, opened 2022-08-01
Handle conditional files in `post_gen_project.py`


lsorber:
1. We currently only handle conditional directories in `post_gen_project.sh`.
2. We should handle conditional files in `post_gen_project` as well ~~because `%` is not a supported filepath character on Windows~~.
3. We should not assume that bash is available [1]:
> For portability, you should use Python scripts (with extension .py) for your hooks, as these can be run on any platform.

[1] https://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html#using-pre-post-generate-hooks

lsorber:
See also: https://til.simonwillison.net/cookiecutter/conditionally-creating-directories

PR #79, opened 2022-08-02
feat: add Windows support

lsorber:
- Adds support for Windows (both `cruft create` and Dev Containers).
- Also closes https://github.com/radix-ai/poetry-cookiecutter/issues/78.


lsorber:
- Cruft create works on Windows 10
- Dev Container works on Windows 10
image

cookiecutter.json
```diff
@@ -14,5 +14,7 @@
   "continuous_integration": ["GitHub", "GitLab"],
   "docstring_style": ["NumPy", "Google"],
   "private_package_repository_name": "",
-  "private_package_repository_url": ""
+  "private_package_repository_url": "",
+  "__package_name_kebab_case": "{{ cookiecutter.package_name|slugify }}",
+  "__package_name_snake_case": "{{ cookiecutter.package_name|slugify(separator='_') }}"
```
lsorber, line 19: FYI:
1. These are private Cookiecutter parameters [1] – i.e. they are hidden from the user during `cruft create`.
2. We need these because neither `|` nor `"` are legal characters in filepaths on Windows.

[1] https://cookiecutter.readthedocs.io/en/stable/advanced/private_variables.html#private-variables

sinopeus (commented):
How did you get this to work? I get an error on Windows 10:

Screenshot 2022-08-08 110008

Apparently the last remaining Jinja templating (in the template directory's name) still doesn't play well with Windows.

sinopeus:
We discussed this and apparently you need to have a recent version of Cookiecutter installed (I had 1.7.3 which gave issues, after installing 2.11.0 things were fine). All works well now!

sinopeus (approved):

PR #82, opened 2022-08-03
feat: add Linux support

lsorber:
Before this PR, Linux support was not great:
1. Within the Dev Container, you are `root`.
2. All files created by you in the Dev Container are therefore owned by `root` in the container.
3. On Linux, files you create within the Dev Container in a (bind-)mounted volume are unfortunately then also owned by `root` on the host system [1]. The result is that you cannot edit those files later on.

This PR brings Linux support to the same level as macOS by introducing a new user that has the same user id and group id as your user on the host machine. This addresses the ownership issue above, and is also better for security as the application won't run as a privileged user [2].

[1] https://github.com/moby/moby/issues/3206
[2] https://pythonspeed.com/articles/root-capabilities-docker-security/


sinopeus (approved):
Just tested and it works nicely. LGTM!

PR #86, opened 2022-08-08
feat: absolufy imports

### LINKED ISSUE #85: Add absolufy-imports to pre-commit hooks
https://github.com/MarcoGorelli/absolufy-imports

lsorber:
This PR will automatically make imports absolute. PEP8 recommends using absolute imports over relative imports [1]:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
> ```python
> import mypkg.sibling
> from mypkg import sibling
> from mypkg.sibling import example
> ```
> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
> ```python
> from . import sibling
> from .sibling import example
> ```
> Standard library code should avoid complex package layouts and always use absolute imports.

Closes #85.

[1] https://peps.python.org/pep-0008/#imports


sinopeus (approved):
Nice little library! I've always hated relative imports, great to finally see myself vindicated by PEP8 ;-) LGTM!

PR #94, opened 2022-08-14
make commitizen optional

bryant-finney:


lsorber:
Hi @bryant-finney, we'll be adding this feature in #109!

PR #101, opened 2022-08-26
Resolve MD033 from markdown lint

kanya-approve:
https://github.com/DavidAnson/markdownlint/blob/v0.25.1/doc/Rules.md#md033


lsorber:
Hi @kanya-approve, thanks for the PR! I won't be accepting this PR at this time however. I'd like to add support for Markdown lint, but I'd rather not add another file to the root of the repo. This adds complexity for new users because of the implicit dependency on Node.js, and one of our goals is to try and keep this Cookiecutter template as simple as possible.

PR #120, opened 2022-11-15
fix: enable zsh history

### LINKED ISSUE #108: Enable zsh history
With:
```
HISTFILE=~/.zsh_history
HISTSIZE=1000
SAVEHIST=1000
```

lsorber:
Closes #108.


sinopeus (approved):

tanguyNaets:
Would it make sense to map the local `.zsh_history file` in the docker-compose file? That way, our commands history doesn't disappear each time we rebuild the devcontainer without cache. And that would ease reusing commands we used in other projects as well.
Only drawback is that the mapping would fail if a user doesn't already have a zsh_history file on his/her local machine (which is quite unlikely)

In docker-compose.yml volumes section:
```
- ~/.zsh_history:/home/app/.zsh_history
```

lsorber:
> Would it make sense to map the local `.zsh_history file` in the docker-compose file? That way, our commands history doesn't disappear each time we rebuild the devcontainer without cache. And that would ease reusing commands we used in other projects as well.

That would be nice indeed, but I don't think we can count on that file existing in general (we also need to support Windows and Linux, which don't use zsh by default). And if the file doesn't exist, Docker Compose will fail to start the container I believe.

PR #123, opened 2022-11-17
feat: support GitHub Codespaces

lsorber:
Problem: before this PR, GitHub Codespaces starts up, but is not usable because the `dev` service mounts the host's `~/.gitconfig` and `~/.ssh/known_hosts`, which are not available on GitHub Codespaces and so instead you see them mounted as empty directories in the dev container.

Solution: the solution is to have separate `devcontainer` service in the `docker-compose.yml` that minimally configures the service for a real Dev Container, and then to have a `dev` service that extends the `devcontainer` service and adds additional configuration that is necessary if you run it as a regular container without the Dev Container conveniences (such as .gitconfig, known_hosts, and forwarding the host's SSH agent).

### DIFF
--- a/{{ cookiecutter.__package_name_kebab_case }}/.devcontainer/devcontainer.json
+++ b/{{ cookiecutter.__package_name_kebab_case }}/.devcontainer/devcontainer.json
@@ -1,7 +1,7 @@
 {
     "name": "{{ cookiecutter.__package_name_kebab_case }}",
     "dockerComposeFile": "../docker-compose.yml",
-    "service": "dev",
+    "service": "devcontainer",
     "workspaceFolder": "/app/",
     "overrideCommand": true,
     "postStartCommand": "cp --update /opt/build/poetry/poetry.lock /app/ && mkdir -p /app/.git/hooks/ && cp --update /opt/build/git/* /app/.git/hooks/",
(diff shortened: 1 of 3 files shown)


{{ cookiecutter.__package_name_kebab_case }}/.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "{{ cookiecutter.__package_name_kebab_case }}",
     "dockerComposeFile": "../docker-compose.yml",
-    "service": "dev",
+    "service": "devcontainer",
```
sinopeus, line 4: Does this change mean that on VS Code we'll use the stripped down `devcontainer` profile as well? Wouldn't that break e.g. the `~/.gitconfig` mount?

lsorber: Yes, that's right, VS Code (and any other Dev Container compatible tooling such as GitHub Codespaces or the Dev Container CLI) will use the `devcontainer` service.

With regard to the `~/.gitconfig` file, I checked and Dev Containers add those conveniences for us automatically. It's only if we don't use Dev Containers but vanilla Docker (Compose) that we need to take care of the things that Dev Containers normally takes care of (like `.gitconfig`, but also forwarding the SSH agent).

PR #130, opened 2022-11-24
fix: remove python-no-eval hook

### LINKED ISSUE #126: `python-no-eval` hook generates false positives when using torch
https://github.com/pre-commit/pygrep-hooks/issues/128

lsorber:
This PR removes the `python-no-eval` hook because of false positives caused by PyTorch [1].

Closes #126.

[1]  https://github.com/pre-commit/pygrep-hooks/pull/103


RubenPants (approved):
LGTM!

PR #142, opened 2022-12-16
fix: fix directory permissions

lsorber:
Problem: `/home/app/.cache/pypoetry/` was owned by `root` instead of the `app` user. Probably caused by BuildKit mounting the cache folder with the wrong user.

Solution: create `/home/app/.cache/pypoetry/` and `/home/app/.config/pypoetry/` with the `app` user before letting BuildKit mount any cache folders (or secrets).


RubenPants (approved):
Fixes all problems on our project. Thanks! :)

PR #156, opened 2023-01-08
feat: improve ruff config

lsorber:
Changes:
1. Upgrade Ruff to 0.0.215.
2. Ignore RET504: the idea is to remove unnecessary assignments before the return statement. However, we shouldn't be too strict about this as sometimes you do want to add an assignment for clarity.
3. Mark F401 as unfixable: we don't want to auto remove unused imports if you're importing them for the first time. Ruff will continue to mark unused imports as linting issues though.

### DIFF
--- a/{{ cookiecutter.__package_name_kebab_case }}/.pre-commit-config.yaml
+++ b/{{ cookiecutter.__package_name_kebab_case }}/.pre-commit-config.yaml
@@ -65,7 +65,7 @@ repos:
       - id: ruff
         name: ruff
         entry: ruff
-        args: [--fixable F401{% if cookiecutter.development_environment == "simple" %}, --fix-only{% endif %}]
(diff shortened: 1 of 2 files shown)


RubenPants:
For "Mark F401 as unfixable", could we still allow this to happen when running `poe lint`, but prevent it from happening on save? E.g. by updating `devcontainer.json`?

lsorber:
> For "Mark F401 as unfixable", could we still allow this to happen when running `poe lint`, but prevent it from happening on save? E.g. by updating `devcontainer.json`?

Yes, good idea. I think we can add `--fixable F401` to the pre-commit config so that it does autofix on commit, but not while you are editing a file.

RubenPants (approved):
Fixed small bug, looks good now!

lsorber:
Thanks @RubenPants!

PR #157, opened 2023-01-08
feat: add up/down arrow cycling of command history

lsorber:
Implements https://github.com/zsh-users/zsh-autosuggestions/issues/646#issuecomment-1011417964.


RubenPants:
Still doesn't seem to work?

https://user-images.githubusercontent.com/21692787/211207802-56fb98c6-11a3-45e1-a180-db5955473b4c.mov

lsorber:
Strange, it should work according to [1] [2]. Did you rebuild the container after updating the Dockerfile? You can also modify `.zshrc` and then source it with `source ~/.zshrc`.

[1] https://unix.stackexchange.com/questions/97843/how-can-i-search-history-with-text-already-entered-at-the-prompt-in-zsh
[2] https://web.cs.elte.hu/local/texinfo/zsh/zsh_28.html

RubenPants:
Yes, the Dockerfile was rebuilt. `cat ~/.zshrc` and `source ~/.zshrc` return what's expected (but the problem stays).
image

lsorber:
Replaced `zsh-autosuggestions` with `zsh-history-substring-search` [1], which should improve the experience.

[1] https://github.com/zsh-users/zsh-history-substring-search

lsorber:
Superseded by #176.

PR #161, opened 2023-01-11
fix: simplify and improve ruff config

lsorber:
ERA001 only applies in strict mode, but when it does it's preferable not to have commented code removed on save.


RubenPants (approved):
LGTM!

PR #162, opened 2023-01-20
Copy files in app image as app user/group

RubenPants:


{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -119,7 +119,7 @@ RUN ln -s /run/secrets/poetry-auth /home/app/.config/pypoetry/auth.toml
 FROM base AS app
 
 # Copy the package source code to the working directory.
-COPY . .
+COPY --chown=app:app . .
```
lsorber, line 122: Could you also add this change to the other `COPY` commands (I believe there are a few others above)? Thanks!

lsorber (approved):
LGTM! Let's get this fixed.

PR #170, opened 2023-02-08
fix: let Dev Containers manage Docker mounts

lsorber:
This removes the bind mount from the `devcontainer` Docker Compose service so that the Dev Containers CLI can manage the mounts as it wants to.

Previously, starting a Dev Container by cloning a repo into a (named) Docker volume would lead to a crash. After this PR, this will work as intended. Which also means that the `Open in Dev Containers` badge actually works!


RubenPants:
@lsorber the pipeline seems to fail?

lsorber:
> @lsorber the pipeline seems to fail?

Will fix!

lsorber:
Superseded by #176.

PR #196, opened 2023-08-23
feat: make Docker image configurable

PGerardi:


{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -1,6 +1,10 @@
 # syntax=docker/dockerfile:1
 ARG PYTHON_VERSION={{ cookiecutter.python_version }}
+{%- if cookiecutter.processing_unit == "GPU"%}
+FROM radixai/python-gpu:cuda11.8-python$PYTHON_VERSION AS base
+{%- else %}
 FROM python:$PYTHON_VERSION-slim AS base
+{%- endif %}
```
lsorber, line 7: In line with the previous comment, this would become:
```suggestion
FROM {{ cookiecutter.docker_image }} AS base
```

.github/workflows/test.yml
```diff
@@ -31,13 +31,12 @@ jobs:
       - name: Scaffold Python package
         run: |
           pip install --no-input cruft
-          cruft create --no-input --extra-context '{"package_name": "My Package", "python_version": "3.8", "with_fastapi_api": "1", "with_typer_cli": "1"}' ./template/
+          cruft create --no-input --extra-context '{"package_name": "My Package", "python_version": "3.8", "processing_unit": "GPU", "with_fastapi_api": "1", "with_typer_cli": "1"}' ./template/
```
lsorber, line 34:
```suggestion
          cruft create --no-input --extra-context '{"package_name": "My Package", "python_version": "3.8", "docker_image": "radixai/python-gpu:cuda11.8-python$PYTHON_VERSION", "with_fastapi_api": "1", "with_typer_cli": "1"}' ./template/
```

For the Docker image tag, I was thinking it might be nice to follow the official Python tag structure, which looks like `python:3.8-slim`. In our case, the GPU-supported version could then be `radixai/python-gpu:3.8-cuda11.8`?

lsorber (changes requested):

sinopeus:
The idea is good but the fact we would use a custom Docker image is the main weakness here, I think. That ideally needs to be built for every new Python version + for every architecture which is a serious investment. Doing that for a subset of architectures/versions is also fine but needs to be made explicit.

I've also also looked at the Dockerfile at https://github.com/radix-ai/python-gpu/blob/main/Dockerfile and given the Docker build parameters I would expect that to look different (e.g. you have `ENV CUDA=11.8` but are not using that variable when installing the dependencies, you always install 11.8). I would maybe clean the part where you install CUDA up and move it into a script that we include in the Cookiecutter, and that we only add in the scaffolded project depending on whether we want GPU support. But then you have the issue of e.g. compatibility issues between certain versions of Python and CUDA.

PGerardi:
@sinopeus 

I see your point. But the thing is that the compatibility between the python, CUDA version, CuDNN, TensorRT should be tested before used. If we put it in a script I think the main advantage of this feature is removed, which is the reduced set-up time. Now one person should test it, and can adjust the Dockerfile. I agree however with the CUDA version ENV comment, and will clean this.

sinopeus:
To be clear, that script would be run as part of the Docker image build, so
it would come down to the same thing 🙂

PGerardi:
I've added all variables as a build arg, so we can easily support new version releases in the future: https://github.com/radix-ai/python-gpu/blob/main/Dockerfile with the matrix object in the publish.yml.

README.md
```diff
@@ -62,6 +62,7 @@ To update your Python project with the latest template:
...
+| `docker_image`                  <br> "python:$PYTHON_VERSION-slim"                                        | The docker image we want to use a base image. If no cuda support is needed, use python:$PYTHON_VERSION-slim. If cuda support is needed, choose radixai/python-gpu:$PYTHON_VERSION-cuda11.8 or radixai/python-gpu:$PYTHON_VERSION-cuda12.2. IMPORTANT: We support CUDA Toolkit 11.8 and 12.2 with python versions 3.8, 3.9, 3.10, 3.11 and 2.12.                                                                                                                                                                                                                                                                                                                                              |
```
lsorber, line 65: Table formatting and description suggestion:
```suggestion
| `docker_image`                    <br> "python:$PYTHON_VERSION-slim"                | The base Docker image to use for the Dev Container and application. The $PYTHON_VERSION build argument is equal to the `python_version` value by default, but may be overridden when building the image to test different Python versions. If CUDA support is required, you may use radixai/python-gpu:$PYTHON_VERSION-cuda11.8 or radixai/python-gpu:$PYTHON_VERSION-cuda12.2.                   |
```

lsorber (changes requested):

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -50,7 +50,6 @@ RUN --mount=type=cache,uid=$UID,gid=$GID,target=/home/user/.cache/pypoetry/ \
     poetry install --only main --no-interaction
 
 
-
 FROM base as ci
```
lsorber, line 53:
```suggestion

FROM base as ci
```

lsorber (approved):

PR #205, opened 2023-11-20
Reduce app image size

sinopeus:


{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -39,6 +31,27 @@ ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
...
+    $POETRY_VIRTUAL_ENV/bin/pip install poetry~=$POETRY_VERSION
+RUN ln -s $POETRY_VIRTUAL_ENV/bin/poetry /usr/bin/poetry
+
+# Install compilers that may be required for certain packages or platforms.
+RUN --mount=type=cache,target=/var/cache/apt/ \
+    --mount=type=cache,target=/var/lib/apt/ \
+    apt-get update && \
+    apt-get install --no-install-recommends --yes build-essential
+
+USER user
```
lsorber, line 53: I think the risk of things going wrong will be less if the venv is installed by the non-root user. Otherwise, we could run into permission issues later on when running the image with the non-root user.

This suggestion moves the `USER root` command down the apt-get install section.
```suggestion
# Install Poetry in separate venv so it doesn't pollute the main venv.
ENV POETRY_VERSION 1.6.1
ENV POETRY_VIRTUAL_ENV /opt/poetry-env
RUN --mount=type=cache,target=/root/.cache/pip/ \
    python3 -m venv $POETRY_VIRTUAL_ENV && \
    $POETRY_VIRTUAL_ENV/bin/pip install -U pip setuptools && \
    $POETRY_VIRTUAL_ENV/bin/pip install poetry~=$POETRY_VERSION
RUN ln -s $POETRY_VIRTUAL_ENV/bin/poetry /usr/bin/poetry

# Install compilers that may be required for certain packages or platforms.
USER root
RUN --mount=type=cache,target=/var/cache/apt/ \
    --mount=type=cache,target=/var/lib/apt/ \
    apt-get update && \
    apt-get install --no-install-recommends --yes build-essential
USER user
```

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -1,6 +1,10 @@
 # syntax=docker/dockerfile:1
 ARG PYTHON_VERSION={{ cookiecutter.python_version }}
 FROM {{ cookiecutter.docker_image }} AS base
+
+# Remove docker-clean so we can keep the apt cache in Docker build cache.
+RUN rm /etc/apt/apt.conf.d/docker-clean
```
lsorber, line 6: Good to add this indeed, otherwise the apt-get cache isn't very useful.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -51,7 +64,7 @@ RUN --mount=type=cache,uid=$UID,gid=$GID,target=/home/user/.cache/pypoetry/ \
 
 
 
-FROM base as ci
+FROM builder as ci
 
 # Allow CI to run as root.
 USER root
```
lsorber, line 70: I don't like these two lines. I believe we added them for compatibility with GitLab CI, but the trouble is that we're testing in an environment that differs from the environment in which we're running the app (`ci` uses the root user while `app` uses the non-root user). Do you think this is fixable in this PR, or should we address that in a separate one?

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -124,6 +137,9 @@ RUN ln -s /run/secrets/poetry-auth /home/user/.config/pypoetry/auth.toml
 
 FROM base AS app
 
+# Copy the virtual environment from the builder stage
```
lsorber, line 140:
```suggestion
# Copy the virtual environment from the builder stage.
```

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -124,6 +137,9 @@ RUN ln -s /run/secrets/poetry-auth /home/user/.config/pypoetry/auth.toml
 
 FROM base AS app
```
lsorber, line 138: I'm a little concerned about this: if we do this we'll miss out on `build-essential`, which comes with compilers but also shared libraries. If the compilers were used to compile certain packages, won't the `app` stage break without the shared libraries that those compiled packages depend on?

{{ cookiecutter.__package_name_kebab_case }}/.dockerignore
```diff
@@ -0,0 +1 @@
+.git
```
lsorber, line 1: For my understanding, what is the benefit of this? As I understand it this will avoid copying the `.git` folder into the Docker context which is available during the build process. But we don't actually copy this directory from the context in to the Docker image, do we? I'd prefer not to add files to the repo root if possible unless there's a clear benefit to doing so.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -51,7 +64,7 @@ RUN --mount=type=cache,uid=$UID,gid=$GID,target=/home/user/.cache/pypoetry/ \
 
 
 
-FROM base as ci
```
lsorber, line 54: I'd prefer if we can bring the `ci` image as close to the `app` image as possible. Right now there are some pretty big changes between the two, and so something that works on the CI may not work at all when you actually build and deploy it. Can this also be based off of `base` like `app` is?

(Before answering this, it's probably better to see my other comment about the base image of the `app` stage.)

lsorber (commented):
Thanks for the PR @sinopeus! Mostly small comments, but one big one about the base image for the `app` and `ci` stages that we need to iron out first.

{{ cookiecutter.__package_name_kebab_case }}/.dockerignore
sinopeus: This line will copy everything including `.git`: https://github.com/radix-ai/poetry-cookiecutter/blob/4df178fcaf036ac89387bfde546b23104466385b/%7B%7B%20cookiecutter.__package_name_kebab_case%20%7D%7D/Dockerfile#L144. So I think this is a necessary change.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: Yes, that was my concern at first as well, but it simply worked even with things like NumPy installed which depend on the C shared libraries. I think the reason everything works is the following. [`build-essential` is a Debian meta-package](https://packages.ubuntu.com/focal/build-essential) which apart from libc installs a C and C++ compiler, `make` and the Debian package development tools. The only thing you actually need in the app image is libc - the compilers etc are not needed at runtime. Looking a little closer [this stage](https://hub.docker.com/layers/library/python/3.9.18-slim-bullseye/images/sha256-ac85017938f7663a8ce2e1679b8512176c15622fa70d37a33043e3333168d787?context=explore) in the official Python Docker image, we can see that libc is installed in the official Python Docker images. So at least for build-essential, there is no risk of missing shared libraries or the like. If there are other shared libraries we're compiling against those should of course be installed in the `base` layer but that's a specific use case.

sinopeus:
I tried installing the Poetry env as a non-root user, but I was relying on symlinking to `/usr/bin/poetry` to avoid juggling two virtualenvs, and creating that symlink of course crashes if you're not `root`. So I reverted that change - the discussion got lost though.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: This is not possible, I think - you'll need Poetry and `build-essentials` in the `ci` image. I do understand the concern but I think there's no real risk there if you agree with [my response to the comment regarding the exclusion of `build-essential` from the `app` image](https://github.com/radix-ai/poetry-cookiecutter/pull/205#discussion_r1401983231).

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: I agree that this is a pain. And I honestly don't even remember what the exact GitLab CI bug was that we fixed with this 🙂 I'd add it to the issue list and fix it later though.

lsorber: OK, sure!

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
lsorber: Excellent, thank you for the analysis!

{{ cookiecutter.__package_name_kebab_case }}/.dockerignore
lsorber: Makes sense.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
lsorber: Ignore this comment, I missed something important when I posted it.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -39,6 +31,27 @@ ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
...
 
+FROM base as builder
+
+USER root
+
+# Install Poetry in separate venv so it doesn't pollute the main venv.
+ENV POETRY_VERSION 1.6.1
+ENV POETRY_VIRTUAL_ENV /opt/poetry-env
+RUN --mount=type=cache,target=/root/.cache/pip/ \
+    python3 -m venv $POETRY_VIRTUAL_ENV && \
```
lsorber, line 42: Could you maybe also optimise lines 27-29 like this, in line with your style here:
```python
ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
ENV PATH $VIRTUAL_ENV/bin:$PATH
RUN python -m venv $VIRTUAL_ENV
```

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -39,6 +31,27 @@ ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
...
+    $POETRY_VIRTUAL_ENV/bin/pip install -U pip setuptools && \
+    $POETRY_VIRTUAL_ENV/bin/pip install poetry~=$POETRY_VERSION
+RUN ln -s $POETRY_VIRTUAL_ENV/bin/poetry /usr/bin/poetry
```
lsorber, line 45: I suggest to install this to `/usr/local/bin/poetry` instead, which is for user-managed packages (whereas `/usr/bin/` is for system-managed packages):
```suggestion
RUN ln -s $POETRY_VIRTUAL_ENV/bin/poetry /usr/local/bin/poetry
```

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -39,6 +31,27 @@ ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
 # Set the working directory.
 WORKDIR /workspaces/{{ cookiecutter.__package_name_kebab_case }}/
 
+FROM base as builder
```
lsorber, line 34: Based on my review, I think it might be more clear to call this stage `poetry`, as that's what defines this stage to me: it makes Poetry available to install (and compile, if needed) packages. Would that be OK, or do you prefer a different name?

lsorber (commented):
Another review round, we're getting closer!

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
lsorber: I'm more or less convinced by your answer, but I still have an inkling there's something else I'm missing. Let me get back to you on this!

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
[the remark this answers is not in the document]
sinopeus: Fixed!

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -39,6 +31,27 @@ ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
...
+
+USER root
+
+# Install Poetry in separate venv so it doesn't pollute the main venv.
+ENV POETRY_VERSION 1.6.1
+ENV POETRY_VIRTUAL_ENV /opt/poetry-env
+RUN --mount=type=cache,target=/root/.cache/pip/ \
+    python3 -m venv $POETRY_VIRTUAL_ENV && \
+    $POETRY_VIRTUAL_ENV/bin/pip install -U pip setuptools && \
```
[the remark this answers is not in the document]
sinopeus: Fixed

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: Fixed

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: Fully agree! Fixed

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
sinopeus: Maybe the fault in my reasoning will come to you in a dream, or something of the sort! Do let me know if anything pops up.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -32,13 +24,33 @@ RUN groupadd --gid $GID user && \
...
-RUN python -m venv /opt/{{ cookiecutter.__package_name_kebab_case }}-env
-ENV PATH /opt/{{ cookiecutter.__package_name_kebab_case }}-env/bin:$PATH
 ENV VIRTUAL_ENV /opt/{{ cookiecutter.__package_name_kebab_case }}-env
+ENV PATH $VIRTUAL_ENV/bin:$PATH
+RUN python -m venv $VIRTUAL_ENV
 
 # Set the working directory.
 WORKDIR /workspaces/{{ cookiecutter.__package_name_kebab_case }}/
 
+FROM base as poetry
```
lsorber, line 34: I just remembered we separated each stage by 3 blank lines (once suggested by Ruben to help see when a new stage begins):
```suggestion


FROM base as poetry
```

sinopeus: Fixed

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -1,6 +1,10 @@
 # syntax=docker/dockerfile:1
 ARG PYTHON_VERSION={{ cookiecutter.python_version }}
 FROM {{ cookiecutter.docker_image }} AS base
+
+# Remove docker-clean so we can keep the apt cache in Docker build cache.
+RUN rm /etc/apt/apt.conf.d/docker-clean
+
```
lsorber, line 7: One newline too many 🙈!

sinopeus: Fixed

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -32,13 +24,35 @@ RUN groupadd --gid $GID user && \
...
+FROM base as poetry
+
+USER root
+
+# Install Poetry in separate venv so it doesn't pollute the main venv.
+ENV POETRY_VERSION 1.6.1
+ENV POETRY_VIRTUAL_ENV /opt/poetry-env
+RUN --mount=type=cache,target=/root/.cache/pip/ \
+    python -m venv $POETRY_VIRTUAL_ENV && \
+    $POETRY_VIRTUAL_ENV/bin/pip install poetry~=$POETRY_VERSION && \
```
lsorber, line 44: Are we sure this will actually install it in the `$POETRY_VIRTUAL_ENV`?

sinopeus: Just double-checked and it does 🙂

sinopeus: (Otherwise the build would have failed)

lsorber: ChatGPT says that calling pip this way will in fact install Poetry to `$POETRY_VIRTUAL_ENV` 👌.

{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
lsorber: I'm satisfied!

PR #206, opened 2023-11-29
feat: update GitLab CI/CD to use the Dev Container CLI

### LINKED ISSUE #193: Refactor GitLab CI to use the Dev Container CLI

lsorber:
Closes #193.

### DIFF
--- a/{{ cookiecutter.__package_name_kebab_case }}/.gitlab-ci.yml
+++ b/{{ cookiecutter.__package_name_kebab_case }}/.gitlab-ci.yml
@@ -11,6 +11,15 @@ variables:
     matrix:
       - PYTHON_VERSION: ["{{ cookiecutter.python_version }}"]
 
+.install_devcontainers_cli: &install_devcontainers_cli
+  cache:
+    paths:
+      - .apk_cache
+      - .npm_cache
+  before_script:
+    # Install the Dev Container CLI.
+    - mkdir -p .apk_cache && apk add --update-cache --cache-dir .apk_cache nodejs npm
+    - npm install --global --cache .npm_cache --prefer-offline @devcontainers/cli@0.54.0
 
 # Build the Dev Container.
 Build:
@@ -23,11 +32,7 @@ Build:
     - |
       # Log in to the Docker registry.
       echo "$CI_REGISTRY_PASSWORD" | docker login --username "$CI_REGISTRY_USER" --password-stdin "$CI_REGISTRY"
-      
-      # Install the Dev Container CLI.
-      apk add --update npm
-      npm install --location=global @devcontainers/cli@0.54.0
-      
+ 
       # Compute a hash for the Dev Container image.
       export CI_IMAGE_SHA="$(sha1sum Dockerfile poetry.lock pyproject.toml | sha1sum | cut -c 1-8)"
       echo "CI_IMAGE_SHA=$CI_IMAGE_SHA" >> .env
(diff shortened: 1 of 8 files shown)


{{ cookiecutter.__package_name_kebab_case }}/Dockerfile
```diff
@@ -64,26 +64,6 @@ RUN --mount=type=cache,uid=$UID,gid=$GID,target=/home/user/.cache/pypoetry/ \
 
 
 
-FROM poetry as ci
```
sinopeus, line 67: 🎉

sinopeus:
This looks like a good change that adds more consistency which I always like 🙂 I'll give it a spin later today!

{{ cookiecutter.__package_name_kebab_case }}/.gitlab-ci.yml
```diff
@@ -1,103 +1,74 @@
...
-  artifacts:
-    reports:
-      dotenv: .env
+variables:
+  DOCKER_TLS_CERTDIR: "/certs"
+
+.parallel:
+  parallel:
+    matrix:
+      - PYTHON_VERSION: [{{ cookiecutter.python_version }}]
```
sinopeus, line 12: We need to add quote marks to avoid [this issue](https://gitlab.com/gitlab-org/gitlab/-/issues/349209):

```suggestion
      - PYTHON_VERSION: ["{{ cookiecutter.python_version }}"]
```

{{ cookiecutter.__package_name_kebab_case }}/.gitlab-ci.yml
```diff
@@ -1,103 +1,74 @@
...
 
-# Base Docker build script.
-.docker:
-  image: docker:stable
+# Build the Dev Container.
```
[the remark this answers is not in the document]
lsorber: Great, thank you! I took what you did here and added it as an additional `extends:` so we don't have to rely on a YAML anchor reference.

{{ cookiecutter.__package_name_kebab_case }}/.gitlab-ci.yml
```diff
@@ -1,103 +1,81 @@
...
-# Base Docker build script.
-.docker:
-  image: docker:stable
+.install_devcontainers_cli:
+  cache:
+    paths:
+      - .apk_cache
+      - .npm_cache
+  before_script:
+    - mkdir -p .apk_cache && apk add --cache-dir .apk_cache npm
```
sinopeus, line 20: I don't think we should remove the `--update-cache` flag here, otherwise this will fail the first time we install it as the APK cache won't be up to date in the project cache. I suppose you want to avoid fetching the APK package list. What we could do is something like this:


```suggestion
    - mkdir -p .apk_cache && (apk add --cache-dir .apk_cache npm || apk add --update-cache --cache-dir .apk_cache npm)
```

But that may be overkill 🙂

lsorber: I actually tried running `mkdir -p .apk_cache && apk add --cache-dir .apk_cache npm` in a fresh Docker container before making this change and it ran successfully. Am I missing something?

sinopeus: Weird, I could swear I've been bitten by that before, but it just works so that's great!

PR #207, opened 2023-12-11
fix: Skip no-commit-to-branch hook in CI

sinopeus:
The `no-commit-to-branch` hook fails on merge commits on `main` when using GitHub Actions. That specific hook can be skipped using the `SKIP` environment variable. We can do that in the `poe lint` task definition based on whether we are in CI or not. To know that, we can check for the existence of the `CI` environment variable and if it exists, set `SKIP=no-commit-to-branch` when running `pre-commit`.

### DIFF
--- a/{{ cookiecutter.__package_name_kebab_case }}/docker-compose.yml
+++ b/{{ cookiecutter.__package_name_kebab_case }}/docker-compose.yml
@@ -17,6 +17,7 @@ services:
     {%- if not cookiecutter.private_package_repository_name %}
     environment:
       - POETRY_PYPI_TOKEN_PYPI
+      - CI
     {%- else %}
     secrets:
       - poetry-auth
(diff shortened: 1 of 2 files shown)


sinopeus:
I'll add that the way I had to fix this (specifically exposing the `CI` env variable to the dev container) shows that the approach of using devcontainers/cli for CI has some downsides, like having to explicitly expose environment variables in `docker-compose.yml` instead of automatically loading the environment variables. This is annoying in CI as there can be e.g. predefined by the CI platform, defined in an earlier stage of the CI pipeline, ... that you want to use.

sinopeus:
By the way, the alternative here is simply removing the `no-commit-to-branch` pre-commit hook altogether and using protected branches on GitHub/GitLab instead. It's definitely a simpler solution 🙂

lsorber:
Thanks for the PR @sinopeus!

I took the liberty of adding a commit with two small changes:
- Only export `CI` in strict mode, since we have no need for it in simple mode.
- Turn the SKIP into a one-liner.

That said, I'm not sure the complexity of the solution is worth the value that the `no-commit-to-branch` hook brings. I'm actually more inclined to just remove the hook entirely and have this be solved with branch rules as you also suggest.

If you prefer to keep it, we can continue with this PR, but otherwise I'd suggest that we remove it.

sinopeus:
All right, let's open a new one where we simply remove the `no-commit-to-branch` hook 🙂

PR #209, opened 2023-12-19
chore: remove `no-commit-to-branch` hook

sinopeus:
See PR #207 for the reason behind removing this hook.


lsorber:
Thanks for the quick PR! I renamed the type from `chore:` to `fix:` because I believe this addresses an issue you pointed out (as opposed to being a routine task).

PR #216, opened 2024-02-22
feat!: replace Jupyter Lab with VS Code notebook support

JWuzyk:
Improves the experience of using notebooks in VSCode by performing the required setup tasks by default:
- Including the `ms-toolsai.jupyter` extension.
- Including the `ipykernel` package as a dev dependency.


lsorber (changes requested):
Hi @JWuzyk, thanks for this PR! I agree we can improve the experience here. In its current state, isn't this PR more or less equivalent with enabling the `with_jupyter_lab` option?

I would prefer to integrate this in the existing Jupyter feature, or replace the existing Jupyter feature entirely, or remove the existing Jupyter feature entirely.

The replacement option could look something like this:
- change the `with_jupyter_lab` feature to `with_jupyter`
- replace `jupyterlab` with `ipykernel`
- possibly have this new feature enabled by default

The removal option would mean that we keep this PR's changes and remove all `with_jupyter_lab` blocks.

lsorber:
@JWuzyk could you also apply [this trick/workaround to automatically select the right kernel](https://github.com/microsoft/vscode-jupyter/discussions/13032#discussioncomment-8049433) in the `"settings"` section in `devcontainer.json`:
```json
"jupyter.kernels.excludePythonEnvironments": ["/usr/local/bin/python"]
```

We should probably log this as an issue to improve once the Jupyter extension allows you to configure a default kernel explicitly.

JWuzyk:
Added to the PR:
- Enabled format on save for notebooks by default.
- Removed `with_jupyter_lab` option
- Added work around to automatically select correct kernel

lsorber (approved):
Looks like a nice improvement to me! Do you want to check with anybody else before we merge this?

sinopeus (approved):

sinopeus:
The only thing I would add is that these changes are VS Code-centric, people using other IDEs or the CLI will have to install JupyterLab themselves. But I think that's OK so all good for me 🙂

lsorber:
Merged! And renamed as the scope changed slightly.

PR #220, opened 2024-04-02
feat: optimize Dependabot config

lsorber:
Before this PR: the Dependabot configuration is suboptimal because it doesn't differentiate well between applications and packages, and because it doesn't group updates to reduce the number of PRs.

After this PR:
- For Python applications and packages:
  - Every month, Dependabot will create *a single* PR for all direct development dependency updates. This PR will bump the minimum version of these dependencies in `pyproject.toml`.
  - ~~Every month, Dependabot will create *a single* PR for all transitive dependency updates (of both production & development dependencies). This PR only updates the lockfile.~~
- For Python applications:
  - Every month, Dependabot will create *a single*  PR for all direct production dependency updates. This PR will bump the minimum version of these dependencies in `pyproject.toml`.
- For GitHub Actions:
  - The current CI dependencies are updated to the latest versions.
  - Dependabot will create *a single* PR for all CI dependency updates per month.

EDIT: Dependabot unfortunately has a number of limitations currently that prevent us from grouping transitive dependency updates and using a different versioning strategy for them. I've updated the scope of this PR's changes above.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -25,7 +25,7 @@ jobs:
           path: template
 
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: "3.9"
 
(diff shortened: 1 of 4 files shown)


sinopeus (approved):

sinopeus (commented):
LGTM but @lsorber will test this first in the demo package.

lsorber:
> LGTM but @lsorber will test this first in the demo package.

After testing, I encountered some limitations on the way Dependabot can be configured. I updated this PR to conform with those limitations and also updated the changes in the PR's description above.

PR #221, opened 2024-04-02
Adding an option to include ZScaler Root Cert

soxguy:
Devcontainer won't be able to update and pull packages with ZScaler turned on unless their Root CA is trusted.


soxguy:
Thought this was on my local fork.  Disregard.

PR #252, opened 2024-09-24
fix: Conflicting dependencies between ipython and commitizen

r-dh:
The earlier addition of `ipython = ">=8.23.0"` conflicts with commitizen and prevents the Docker image to build successfully.

 - `commitizen` requires `questionary (>=2.0,<3.0)`.
 - `questionary (2.0.0 and 2.0.1)` depend on `prompt_toolkit (>=2.0,<=3.0.36)`.
  - `ipython (>=8.23.0)` depends on `prompt_toolkit (>=3.0.41,<3.1.0)`.

**Conflict:** `prompt_toolkit (>=3.0.41,<3.1.0)` required by `ipython (>=8.23.0)` is incompatible with `prompt_toolkit (<=3.0.36)` required by `questionary` (and thus `commitizen`).

I found that `ipython` version  `>=8.18.0` is compatible with commitizen.

PR #254, opened 2024-10-06
fix: fix prompt-toolkit dependency conflict

lsorber:
Fix the dependency conflict described in https://github.com/commitizen-tools/commitizen/issues/1249 by relaxing the minimum ipython version.


r-dh:
Hey, dropping in to mention that PR #252 is fixing the same issue.

lsorber:
Thank you for letting us know @r-dh, I totally missed that, apologies!

PR #260, opened 2024-12-20
fix: fix casing of FROM AS in Dockerfile

lsorber:
Docker appears to balk at the inconsistent casing of `FROM ... AS` in the `Dockerfile`, and this PR fixes that.


lsorber:
Going to go ahead and merge this because its pretty urgent.

Issue #256, opened 2024-11-04
PyCharm support for DevContainers not working as expected


undo76:
PyCharm has added support for devcontainers, but for some reason it is not working as expected. Some issues:
- libraries are not recognised in the editor
- There is "Connect" option to "devcontainer", but not "dev"

lsorber:
I tested PyCharm's Dev Containers feature on the [Conformal Tights](https://github.com/superlinear-ai/conformal-tights) repo, and it did work for me. You do need to be careful to:
    1. select with Mount Sources, and
    2. configure the Python interpreter to the virtual env created under /opt.

I created #261 to explain exactly what you need to do to get it to work, and also improve the default experience a bit. Let me know what you think!

lsorber:
Unfortunately, it seems that PyCharm does not yet allow to configure the Python interpeter in `devcontainers.json`.

PR #261, opened 2024-12-20
feat: add support for Dev Containers with PyCharm

lsorber:
Changes:
1. Improves the PyCharm Dev Container experience.
2. Include nodejs in the image, which is also needed for the GitHub Copilot PyCharm plugin.
3. Remove the PyCharm Docker Compose interpreter dev environment (which is now replaced by Dev Containers).

Related: #256.


sinopeus (approved):
Tested this and while PyCharm's Dev Containers implementation is itself somewhat janky and certainly not as seamless as VS Code's, this integration works. LGTM!

PR #266, opened 2025-01-16
feat: upgrade to Poetry v2.0

lsorber:
Changes:
- Upgrade to [Poetry v2.0](https://python-poetry.org/blog/announcing-poetry-2.0.0/)
- Upgrade antidote to v1.9.7
- Small improvement to the Ruff config
- Remove codecov

### DIFF
--- a/{{ cookiecutter.__project_name_kebab_case }}/.github/workflows/test.yml
+++ b/{{ cookiecutter.__project_name_kebab_case }}/.github/workflows/test.yml
@@ -40,8 +40,3 @@ jobs:
 
       - name: Test {{ cookiecutter.project_type }}
         run: devcontainer exec --workspace-folder . poe test
-
-      - name: Upload coverage
-        uses: codecov/codecov-action@v5
-        with:
-          files: reports/coverage.xml
--- a/{{ cookiecutter.__project_name_kebab_case }}/Dockerfile
+++ b/{{ cookiecutter.__project_name_kebab_case }}/Dockerfile
@@ -37,7 +37,7 @@ FROM base AS poetry
 USER root
 
 # Install Poetry in separate venv so it doesn't pollute the main venv.
-ENV POETRY_VERSION 1.8.0
+ENV POETRY_VERSION 2.0.1
 ENV POETRY_VIRTUAL_ENV /opt/poetry-env
 RUN --mount=type=cache,target=/root/.cache/pip/ \
     python -m venv $POETRY_VIRTUAL_ENV && \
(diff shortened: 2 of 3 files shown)


{{ cookiecutter.__project_name_kebab_case }}/src/{{ cookiecutter.__project_name_snake_case }}/api.py
sinopeus, line 1: Is there a specific reason why you remove coloredlogs? Doesn't really fit in with this PR.

{{ cookiecutter.__project_name_kebab_case }}/.github/workflows/test.yml
sinopeus, line 1: Is there a specific reason why you remove the codecov upload?

sinopeus (commented):
Left two comments, will still test it on my end as well 🙂

{{ cookiecutter.__project_name_kebab_case }}/src/{{ cookiecutter.__project_name_snake_case }}/api.py
lsorber: As I was upgrading all dependencies, I noticed that this package hasn't been updated in almost 4 years. It felt a bit strange to upgrade everything and keep this package around knowing that isn't maintained any more. I could make it a separate PR, but went with the convenient solution 😅.

{{ cookiecutter.__project_name_kebab_case }}/.github/workflows/test.yml
lsorber: Similar to the previous comment, the trigger to remove it was that I was upgrading all dependencies. IIRC it used to work with zero configuration for open-source projects, but now it requires you to create an account and set an API key. Also, I don't like that we're depending on a commercial service. There may be better solutions: https://github.com/superlinear-ai/poetry-cookiecutter/issues/228.

sinopeus: Makes sense!

{{ cookiecutter.__project_name_kebab_case }}/src/{{ cookiecutter.__project_name_snake_case }}/api.py
sinopeus: I didn't know, out it goes! 🙂

sinopeus:
Everything looks good except one thing that didn't work on my end. I took the liberty of fixing it [in this commit](https://github.com/superlinear-ai/poetry-cookiecutter/pull/266/commits/192d60b57c8ed658fefca6a5f2b0194e99caba6b). Let me know whether you think that makes sense, I find it strange that it worked in CI but not on my machine.

PR #276, opened 2025-02-17
docs: add demo to README

lsorber:
Add the following demo to the README:

https://github.com/user-attachments/assets/28d23137-ebae-47d8-a6e5-11f66abf2a91


sinopeus:
The demo just shows how to scaffold a new project, which is good. But we also used to refer to Conformal Tights as an example project, let's keep that. I'll rescaffold Conformal Tights. I'd also make a more extensive screencast that shows how opening a new project works.

PR #291, opened 2025-02-26
fix: added missing py.typed file in Copier template's src

vanolucas:
The `py.typed` file is present in the cookiecutter template but was missing in the Copier Jinja template.


lsorber:
Hi @vanolucas, the `py.typed` file is actually in the Copier template but only present if you select `typing == "strict"`. But indeed it may make sense to have it even if `typing == "optional"`. Would you like to rename the file so that it's always present? We're also happy to take care of it, either way is fine!

vanolucas:
@lsorber,

Thank you, it makes sense indeed, I missed `{% if with_strict_typing_and_linting %}py.typed{% endif %}.jinja` that already exists in this directory.

However, when I create a new project with the following settings, the `py.typed` file does not get created:

Example 1:

```yml
_commit: v1.2.0
_src_path: gh:superlinear-ai/substrate
author_email: j@j.com
author_name: John
ci: github
project_description: Yo.
project_name: My App
project_type: app
project_url: https://example.com
python_version: '3.13'
typing: strict
with_conventional_commits: true
with_fastapi_api: true
with_typer_cli: true
```

Example 2:

```yml
_commit: v1.2.0
_src_path: gh:superlinear-ai/substrate
author_email: j@j.com
author_name: John
ci: github
project_description: p.
project_name: My App 2
project_type: package
project_url: https://example.com
python_version: '3.13'
typing: strict
with_conventional_commits: true
with_typer_cli: true
```

PR #297, opened 2025-03-08
fix: fix resolution strategy selection

lsorber:
Changes:
1. Make the local environment variables `PYTHON_VERSION` and `RESOLUTION_STRATEGY` available in the container for the `postStartCommand`.
2. Set [Poe the Poet's executor type](https://poethepoet.natn.io/global_options.html#change-the-executor-type) to `simple`. Otherwise, it seemingly reapplies `uv sync` before running a task (related: https://github.com/nat-n/poethepoet/pull/271).


sinopeus (approved):
LGTM!

PR #300, opened 2025-03-11
feat: simplify publishing to PyPI from GitHub Actions

lsorber:
Changes:
1. Simplify the publishing workflow to use the new [Trusted Publisher feature](https://docs.pypi.org/trusted-publishers/adding-a-publisher/). This means you no longer need to create or set a publish token to publish from GitHub Actions.
2. Fix the `runs-on` field and instead install uv, which is not a valid value and [caused the release of Conformal Tights to hang](https://github.com/superlinear-ai/conformal-tights/actions/runs/13792977495).
3. Simplify `docker-compose.yml` by removing the `UV_PUBLISH_TOKEN` unless the project is configured to publish a package without GitHub Actions.

### DIFF
--- a/template/docker-compose.yml.jinja
+++ b/template/docker-compose.yml.jinja
@@ -3,7 +3,7 @@ services:
   devcontainer:
     build:
       target: dev
-    {%- if project_type == 'package' %}
+    {%- if project_type == 'package' and ci != 'github' %}
     environment:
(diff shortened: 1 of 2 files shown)


sinopeus (approved):
Nice change and TIL about Trusted Publisher! LGTM

PR #301, opened 2025-03-12
fix: add missing permission for trusted publishing

lsorber:
Changes:
1. Add missing `id-token: write` permission that is necessary for Trusted Publishing [1].
2. Remove the `UV_PUBLISH_TOKEN` altogether since Trusted Publishing is available for several sources, including GitLab.

[1] https://docs.pypi.org/trusted-publishers/using-a-publisher/


sinopeus (approved):
LGTM

PR #317, opened 2025-09-09
fix: select user-defined minimum Python version in CI test matrix

PGerardi:
Replace hardcoded Python versions [3.10, 3.12] for packages with the user's chosen python_version from copier.yml to ensure compatibility between selected Python version and pyproject.toml requirements.

### DIFF
--- a/template/{% if ci == 'github' %}.github{% endif %}/workflows/test.yml.jinja
+++ b/template/{% if ci == 'github' %}.github{% endif %}/workflows/test.yml.jinja
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


PGerardi:
I fully agree. However, the CI breaks if we select Python 3.12, as it creates a hard requirement in the pyproject.toml file. We need to find a way to reduce that requirement to 3.10 in a parameterized way.

PGerardi:
image

lsorber (approved):
I'm not sure why it was 3.10 and 3.12 originally. The proposed change makes sense as `python_version` is defined as the lowest version that the project is willing to support (for packages). Users can add additional higher Python versions manually after scaffolding the project.

